PR #183, opened 2014-07-08
Agnoster ruby_prompt

rocknrollMarc:
I added a ruby prompt on the right side, I need to fr work :)

### DIFF
--- a/themes/agnoster/fish_right_prompt.fish
+++ b/themes/agnoster/fish_right_prompt.fish
@@ -1,0 +1,3 @@
+function fish_right_prompt
+  rvm-prompt
+end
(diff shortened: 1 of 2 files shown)


rocknrollMarc:
Ah ok get you. Are you sure that a good idea? I have just recently discovered zsh and fish shell and for people who are not that into shell scripting its kind of difficult and they might not know that you can do that. Otherwise Ill do it.

bpinto:
That's why it's important to have a README file, so when they are viewing the theme, they can also read about its functionality.

rocknrollMarc:
Ah ok. Sure ok then Ill do that in the next days as I really like agnoster with the ruby thing. Actualy I think on presto they have added it to agnoster funny

bucaran:
@rocknrollMarc Hi. How is it going?

rocknrollMarc:
LoL yes time flys by doesnt it ill try doing that today got some free time

bucaran:
:+1: Closing until @rocknrollMarc is back.

PR #207, opened 2014-08-20
Fix issue with rbenv plugin where some rbenv commands fail

charlesbjohnson:
Hey there!

rbenv officially supported fish on master some time ago. This just tweaks the rbenv plugin so that it enables shims properly. Previously, commands like `shell` and `rehash` would fail since the plugin was using a workaround for loading shims.

Thanks!


bpinto:
Very good changes! Thanks!

Issue #88, opened 2013-11-11
Dependency between plugins


leira:
To import a dependent plugin from another plugin is useful sometimes, oh-my-zsh has this feature, seems oh-my-fish has not implemented it yet.

bucaran:
#280 and #284 should fix and allow simple plugin interdependency.

#280: Adds `prepend_tree` that basically traverses a source path and prepends each directory matching a list of glob patterns to a destination path (leave out to to add them to `fish_function_path`.

#284: Automatically sources completion and `load` files when `_fish_add_plugin` is called.

_prepend_tree (dirname $argv) # Source dependencies inside current plugin.
_fish_add_plugin # Adds plugin and calls that plugins `plugin.load`.

`_fish_add_plugin` calls `_prepend_tree` behind the scenes which checks whether the path to add already exists in the destination path, saving you the trouble to check for conflicting imports.

bpinto:
There is a new [import](https://github.com/bpinto/oh-my-fish/blob/master/functions/import.fish) function that allows plugins and themes to depend on other plugins.

Ping @leira

PR #284, opened 2015-01-01
Autoloading on plugin add.

bucaran:
> This is a minor commit, but it helps creating more consistent plugin `load` files, it also adds a simple solution for plugin interdependency #88.

This is our current plugin loading strategy:

```fish
# Add all defined plugins
for plugin in $fish_plugins
  _fish_add_plugin $plugin
  _fish_add_completion $plugin
  _fish_source_plugin_load_file $plugin
end
```

If your plugin requires another plugin to work, you should be able to do so inside your `myplugin.load` like:

```fish
_fish_add_plugin plugin
```

This works for simple plugins with only a `plugin.fish` function file, but will fail to source the plugin's load file which is how additional dependencies are defined / required in a per-plugin basis.

`_fish_add_plugin` should automatically source load files and completions.


bucaran:
If anybody knows why two of my past contributions are showing up on this PR, please let me know the reason.

bpinto:
You need to rename your branch with master. Your master branch could also be outdated compared to the upstream one.

bpinto:
I like the simplification, I'm not sure yet if we shouldn't have another function (could be an alias) called require_plugin or something similar. What's your opinion?

My thinking is that it would have a more descriptive name , _functions were created with the goal of being private functions.

bucaran:
I know for sure my branch is up to date with upstream, but let me check again.

That is also an option. This simple change does not create anything new since fish_add_plugin is a core function.

So, do you prefer creating a new function require_tree to add plugins along with completetions and .load sourcing? 

_prepend_tree should be renamed as well since it is not intended to be private.

bpinto:
:+1: 

The doors are always open. :)

PR #286, opened 2015-01-02
[New] Rename `_prepend_tree` to `require_tree` and add `require_plugin` command.

bucaran:
Two commits:
+ `_prepend_tree` should be featured as a public function so it was renamed to `require_tree`.
+ Add `require_plugin` for a simple plugin interdependency solution.

Related #284

### DIFF
--- a/functions/require_plugin.fish
+++ b/functions/require_plugin.fish
@@ -1,0 +1,9 @@
+# Load plugin and dependencies.
+# @params <plugin_name>[..]
+function require_plugin -d "Load a plugin list and dependencies."
+  for plugin in $argv
+    _fish_add_plugin $plugin
+    _fish_add_completion $plugin
+    _fish_source_plugin_load_file $plugin
+  end
+end
--- a/functions/require_tree.fish
+++ b/functions/require_tree.fish
@@ -1,9 +1,8 @@
-#
 # NAME
-#      _prepend_tree - add a dependency tree to a path
+#      require_tree - add a dependency tree to a path
 #
 # SYNOPSIS
-#      _prepend_tree [-P --preview] <source path> [<glob>..]
+#      require_tree [-P --preview] <source path> [<glob>..]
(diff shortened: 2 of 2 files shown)


functions/require_plugin.fish
```diff
@@ -0,0 +1,9 @@
+# Load plugin and dependencies.
+# @params <plugin_name>[..]
+function require_plugin -d "Load a plugin list and dependencies."
+  for plugin in $argv
+    _fish_add_plugin $plugin
```
bpinto: Where are these functions? They could be ported to this function as they don't need to be used outside of this function scope.

bucaran: Inside `oh-my-fish/functions/`.
Good idea, let me double check and the commit.

functions/require_tree.fish
```diff
@@ -74,15 +73,15 @@
 # SEE ALSO
 #      .oh-my-fish/functions/_prepend_path.fish
 #
-# v.0.1.0
+# v.0.1.1
 #/
-function _prepend_tree -d "Load a dependency tree, matching `.fish` files by default."
+function require_tree -d "Load a dependency tree."
```
bpinto: We still add directories with `.fish` files by default. The function description (-d) might not be the best place to add this information, but perhaps on the function description block above?

bucaran: Sure. What do you think about incorporating `require_tree` into the main setup?

bpinto: I'm all in. :+1:

bucaran: OK, let me work on this. By the way,`archlinux` could benefit from this as it is currently using this loading strategy:

```fish
set -l current_dir (dirname $argv)

# add pacman related functions
set fish_function_path $current_dir/pacman $fish_function_path
```

bpinto: Cool.

PR #285, opened 2015-01-01
Add msg plugin

bucaran:
```
   __ _  ___ ___ _
  /  ' \(_-</ _ `/
 /_/_/_/___/\_, /
           /___/ v0.1.0

NAME
     msg - technicolor message printer

SYNOPSIS
     msg [-sn] @<fg:bg> [style]<text>[style]

OPTIONS
     -s
         Do not separate arguments with spaces.
     -n
         Do not output a newline at the end of the message.

STYLES
     _text_                 Bold
     __text__               Underline
     ___text___             Bold and Underline
     /directory/            Directories
     [url]                  Links
     \n                     Line Break
     \t                     Tab Space

     @<fg:bg> fg=bg=RGB|RRGGBB|name|random
       RGB value can be any three to six hex digit or color name.
       e.g, @0fb, @tomato, @random, @error, @ok.

AUTHORS
     @bucaran (Jorge Bucaran)

v. 0.1.0
```

See #283 for more info.

Issue #283, opened 2014-12-25
Should I make this module a plugin?


bucaran:
I made a colored `echo` based command; `msg` for a plugin I am currently working on. It could be useful to other users so I was thinking I should make it a plugin instead. Basically it can do something like this:

```
msg !! Error! -- Well, this is embarrassing...
msg ++ OK -- File ~~ $filename -- downloaded to :/ $directory
msg Visit → :// http://github.com
msg @orchid Dream @chocolate in @cornsilk Technicolor
msg Line1 == Line2 == Line3 ======= Line9
msg @random Only @random God @random Knows
msg @bad Bad @dad Dad @ff0 ":)"
msg _ Underline __ Bold ___ Underline and Bold FTW
```

![sample](https://cloud.githubusercontent.com/assets/8317250/5555028/4817de3e-8ccf-11e4-93d4-3dd57e1f47bc.png)

Here is the `man` page. If you think this is plugin worthy I will submit a PR.

```
    __ _  ___ ___ _
   /  ' \(_-</ _ `/
  /_/_/_/___/\_, /
            /___/ v0.1.0

 NAME
      msg - technicolor message printer

 SYNOPSIS
      msg [-s] [-- -n --normal] [~~ -i --info] [_ -u --under]
          [__ -b --bold] [___] [++ -k --ok --success] [!! -e --error]
          [:// -w --www --url --mail] [:/ -d --dir] [== -l --line-break]
          [=== + \n][@<RRGGBB>] [<text>]

 OPTIONS
      -s
          Should be first if used. Do not separate arguments with spaces.

      -n
          Should be first if used. Do not output a newline at the end of
          the message.

 STYLES
      --   -n --normal              Normal/Default
      ~~   -i --info                Information
      _    -u --under               Underline
      __   -b --bold                Bold
      ___  -bu -ub                  Bold and Underline
      ++   -k --ok --success        Success
      !!   -e --error               Error
      ://  -w --www --url --mail    URLs, Emails, etc
      :/   -d --dir                 Directories
      ==   -l --line-break          Line Break
      ===                           +1 Line Break for extra =

      @<[RGB|RRGGBB|name|random]>
          RGB value can be a three/six hex digit or color name, e.g, @0fb
          @tomato, etc. Use @random to select a random color.

 AUTHORS
      @bucaran (Jorge Bucaran)

 v. 0.1.0
```

PR #287, opened 2015-01-02
[New] Msg plugin / _prepend_tree update.

bucaran:
```
   __ _  ___ ___ _
  /  ' \(_-</ _ `/
 /_/_/_/___/\_, /
           /___/ v0.1.0

NAME
     msg - technicolor message printer

SYNOPSIS
     msg [-sn] @<fg:bg> [style]<text>[style]

OPTIONS
     -s
         Do not separate arguments with spaces.
     -n
         Do not output a newline at the end of the message.

STYLES
     _text_                 Bold
     __text__               Underline
     ___text___             Bold and Underline
     /directory/            Directories
     [url]                  Links
     \n                     Line Break
     \t                     Tab Space

     @<fg:bg> fg=bg=RGB|RRGGBB|name|random
       RGB value can be any three to six hex digit or color name.
       e.g, @0fb, @tomato, @random, @error, @ok.

AUTHORS
     @bucaran (Jorge Bucaran)

v. 0.1.0
```

Some utility functions are declared inside `msg.fish` in order to simplify and separate concerns, like reducing the amount of `sed` calls in the wild. The following provides some documentation for local utilities to `msg`.

### `function --no-scope-shadowing msg._.set.color`
Set style, fg/bg colors and reset the color again. This function modifies variables in the parent scope.
`params [<fg>] [<bg>] [<style>]`

### `function msg._.str.has`
True if substring exists in string. Used to cascade apply `-sn` options when `msg` is called.
`params <substring> <string>`

### `function msg._.str.grab`
Extract string between left and right separators of variable length. Used to grab the part inside the style separators. `_bold_, __underline__, ___bold_and_underline___, [urls], /directories/`
`params <left-sep> [<right-sep>] <string>`

### `function msg._.random.color`
Print a random `RRGGBB` hexadecimal color. 

### `function msg._.get.color`
Translate color names to valid RRGGBB hexadecimal value. This function takes the bulk of the code in the module as it checks for all 256 color names.
`params <color|random>`

See #283 and #285 for more info.

### DIFF
--- a/plugins/msg/msg.fish
+++ b/plugins/msg/msg.fish
@@ -1,0 +1,467 @@
+#    __ _  ___ ___ _
+#   /  ' \(_-</ _ `/
+#  /_/_/_/___/\_, /
+#            /___/ v0.1.0
+#
+# NAME
+#      msg - technicolor message printer
+#
+# SYNOPSIS
+#      msg [-sn] @<fg:bg> [style]<text>[style]
+#
+# OPTIONS
+#      -s
+#          Do not separate arguments with spaces.
+#      -n
+#          Do not output a newline at the end of the message.
+#
+# STYLES
+#      _text_                 Bold
+#      __text__               Underline
+#      ___text___             Bold and Underline
+#      /directory/            Directories
+#      [url]                  Links
+#      \n                     Line Break
+#      \t                     Tab Space
+#
+#      @<fg:bg> fg=bg=RGB|RRGGBB|name|random
+#        RGB value can be any three to six hex digit or color name.
+#        e.g, @0fb, @tomato, @random, @error, @ok.
+#
+# AUTHORS
+#      @bucaran (Jorge Bucaran)
+#
+# v. 0.1.0
+# /
+set --global msg_color_fg   FFFFFF
+set --global msg_color_bg   normal
+set --global msg_style_url  00FF00 $msg_color_bg -u
+set --global msg_style_dir  FFA500 $msg_color_bg -u
+set --global msg_color_err  FF0000
+set --global msg_color_ok   00FA9A
+
+function msg -d "Technicolor printer."
+  set -l reset (set_color normal)(set_color $msg_color_fg)
+  set -l fg $msg_color_fg
+  set -l bg $msg_color_bg
+  # Set style, fg/bg colors and reset. Note this function
+  # modifies variables in the parent scope.
+  # @params [<fg>] [<bg>] [<style>]
+  function --no-scope-shadowing msg._.set.color
+    [ (count $argv) -gt 0 ]
+      and set fg $argv[1]
+    [ (count $argv) -gt 1 ]
+      and set bg $argv[2]
+    # Resets color as a side effect.
+    set_color -b $bg
+    set_color $fg
+    [ (count $argv) -gt 2 ]
+      and set_color $argv[3]
+    set bg $msg_color_bg
+    set fg $msg_color_fg
+  end
+
+  # Default " " whitespace between strings, skip with -s.
+  set -l ws " "
+  # Default \n after message, skip with -n.
+  set argv $argv \n
+  switch $argv[1]
+    case -s\* -n\*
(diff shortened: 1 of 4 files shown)


bucaran:
This PR also shows all my previous commits because the branches were locally merged. This is however a new PR fully in sync with your most recent master.

plugins/msg/msg.fish
```diff
@@ -0,0 +1,467 @@
...
+        echo -en (msg._.set.color)$token$reset
+        [ -z $ws -o $argv[-1] = $token ]
+          or echo -n $space
+    end
+  end
+end
+
+# True if substring exists in string.
+# @params <substring> <string>
+function msg._.str.has
```
bpinto: What is the `_` for?

What do you think about `function _msg_string_has`?

plugins/msg/msg.fish
```diff
@@ -0,0 +1,467 @@
...
+  printf $argv[2] | grep -q $argv[1]
+end
+
+# Extract string between left and right separators of variable length.
+# @params <left-sep> [<right-sep>] <string>
+function msg._.str.grab
```
bpinto: `_msg_string_extract`

plugins/msg/msg.fish
```diff
@@ -0,0 +1,467 @@
...
+# Print a random RRGGBB hexadecimal color.
+function msg._.random.color
+  printf "%02x%02x%02x" (math (random) "%" 255) \
+                        (math (random) "%" 255) \
+                        (math (random) "%" 255)
+end
+
+# Translate color names to valid RRGGBB hexadecimal value.
+# @params <color|random>
+function msg._.get.color
```
bpinto: `_msg_color_parse`

plugins/msg/msg.fish
```diff
@@ -0,0 +1,467 @@
...
+   	case whitesmoke
+      printf F5F5F5
+   	case white
+      printf FFFFFF
+    case "*"
+      printf $argv[1]
+  end
+end
+
+function msg._.test
```
bpinto: When is this used?

plugins/msg/msg.fish
bucaran: I use `_` to mean local functions / utilities.

I really like the `module_name_._.command` convention to name functions given that `.` are valid function names (but not variables) in Fish. I wouldn't mind a lot changing this, but I am writing another plugin right now that is pretty large using this convention consistently and it's really useful when reusing default keywords like `set` or `contains` as they are automatically highlighted by my editor. So I could for instance `msg._.str.contains` and `contains` would be highlighted adding readability without interfering with anything.

bpinto: :+1: 

I don't have a strong opinion on this. Let's try to embrace this convention and see how it goes. :)

plugins/msg/msg.fish
bucaran: I used it just for testing it, it's _not_ strictly necessary.

bpinto: Have you heard about [fish-tank](https://github.com/terlar/fish-tank)? Might be a better solution.

plugins/msg/msg.fish
bucaran: Cool. If you don't mind, I will leave the `._.` as well, as it's shorter than, e.g., `msg.utils.command` which lets me stay under 80 characters per column in a few long one-liners. I

bpinto: :+1:

plugins/msg/msg.fish
bucaran: Yes, but I am not exactly sure how I could test `msg`. For example, there is no `get_color` in Fish so how can I confidently know what color was set by `set_color`? Also, according to `set_color`'s manual, not all terminal emulators support all the color / style features. I will definitely look into this later though.

plugins/msg/msg.fish
bucaran: Thanks.

Issue #96, opened 2013-11-29
Load plugins in background


fazibear:
Hey, 

Where I have lot of plugins I need to wait until they load, what about idea to load it in background ?

for example:

set fish_plugins node brew fry-plugin rake& bundler&

rake and bundler plugins were load in background, after prompt appeared.

Is this possible ?

bucaran:
Here is a little bit of an idea using @bobthecow `tab` plugin:

```
function background -a done_handler
  tab # Open a new terminal tab in iTerm2 or Terminal (creates a new fish session).


  for i in (seq 1 50000)
    echo Loading a bunch of dependencies \#x00$i
  end
  echo OK
  # echo $fish_function_path
  # set -U fish_function_path $fish_function_path \
  #   $fish_path/custom/plugins/mylib

  set -U ___para 100

end
```

bpinto:
We would need to either use universal variables or redo the env changes on the main session. I am afraid this wouldn't be feasible or worth it.

bucaran:
We should close this issue as it seems like something that will not be fixed in the near future or at all.

bpinto:
I guess you are right, as much as I would like to get this done, I don't think this will ever be possible.

Issue #259, opened 2014-11-22
Vundle style plugins


rominf:
It would be great to have the ability to install plugins just like in [vunlde](https://github.com/gmarik/Vundle.vim).

Current approach (all plugins are in the main repo) makes development really slow.

Yes, vunlde's approach has some issues:
1. It should be implemented. I think that it would take 100-200 LOC, because fish is more powerful than vimscript. See https://www.openhub.net/p/vundle
2. Big overhead (one file vs one file + git repo).

But I think that this worth to do.

Related to #88, #96

bucaran:
So in this _open_ model, oh-my-fish would not have more than a very few _out of the box_ plugins and simply provide the habitat where fish would live and thrive with adequate oxygen levels, tolerable temperature, food and hiding places.

bpinto:
At this moment, I don't intend to open plugins to any git url. I don't think it's the right time to have this.

What I think is the best approach is to convert oh-my-fish into a organization and have multiple repositories inside this organization, one for each plugin, and give permission for plugin authors to manage their own plugin repo.

This way, we would just need to change `import` to fetch the files from another repo if they do not exist locally.

bucaran:
:+1: Cool. Let's do it one of these days. We will need a very cool logo for oh-my-fish.

bpinto:
:+1: 

It would be awesome to have a logo for oh-my-fish, but it's beyond my skills to create one. :P

bucaran:
> At this moment, I don't intend to open plugins to any git url. I don't think it's the right time to have this.

:+1: 

> What I think is the best approach is to convert oh-my-fish into a organization and have multiple repositories inside this organization...

We could start creating the organization, migrating the repository and taking it from there. 

I will open a new issue with this so that @bpinto can take a look when he's back.

bpinto:
I have started working on this issue here: https://github.com/bpinto/oh-my-fish/pull/462

bucaran:
:+1: :+1:

bpinto:
The initial implementation #462 was merged to master. Next step is to move all plugins and themes to the [oh-my-fish organization](https://github.com/oh-my-fish).

bpinto:
ping @rominf @samvv

rominf:
I've commented `fish` function. In general, it looks good for the first implementation. Of course, `remove`, `list` and other options are important for future versions.

bpinto:
Thanks! I'm already working on `list`. `remove` should come next.

PR #288, opened 2015-01-04
〈New〉import command.

bucaran:
### Breakdown

+ Rewrites `oh-my-fish.fish` initial configuration and adds in-depth comments. The amount of code is reduced to ~__10__ lines of code.

+ Adds a __new__ [`import`](https://github.com/bucaran/oh-my-fish/blob/import/functions/import.fish) command and introduces the idea of _libraries_. A library is category to  conceptually organize cluster of `.fish` files such as plugins, themes and completions. A plugin intended to be used by other plugins and not by the user of the shell can be thought of as a _library_. 

+ Simplifies `_prepend_tree` reducing amount of code to half.

+  `import` calls [`_prepend_tree`](https://github.com/bucaran/oh-my-fish/blob/import/functions/_prepend_tree.fish) for each plugin in the global `$fish_plugins` prepending subdirectories containing `.fish` files to the `$fish_function_path`. 

Now it's possible to create complex directory trees with `fish` files organized according to the plugin's own API / structure. Paths containing `.fish` files will be added to the path automatically. This should encourage plugin authors to create a consistent project hierarchy without having to worry with load / configuration files modifying the `$fish_function_path` directly. `.load` files can still be used for more pertinent tasks such as downloading assets, declaring global variables, etc. 

> __Note__: `_prepend_tree` runs before `.load` files are sourced, so it is still possible to override any undesired behavior in the `<plugin>.load` file if necessary.

### Notes

`import` replaces `oh_my_fish.fish`, `_fish_add_plugin`, `_fish_add_completion`, `_fish_source_plugin_load_file` functions inside `functions/oh-my-fish.fish/`, adds documentation and calls `_prepend_tree`, but it is still less code than the current configuration. This is possible using Fish `{}` brace expansion:

```fish
    set -l root $fish_path $fish_custom
    # ...
    set -l paths $root/$library $root/$library/completions
    set -l plugins     $paths[1..2]" fish_function_path"
    set -l completions $paths[3..4]" fish_complete_path"
    
    for path in $plugins $completions
      eval "_prepend_path "$path
    end
```

#### Performance
I tested adding _al_ l the existing modules with a macbook pro and while I did notice a small lag when starting the shell the first time, the bottleneck seems to be simply adding so many plugins (and the fact Fish [can not](https://github.com/fish-shell/fish-shell/issues/563) run Fish code in the background) and not `_prepend_tree` running for each plugin, conclusion I reached after commenting it and retesting without being able to notice any significant performance gains. See #96 for more info.

#### `import <path/library>[<path/library>..]`
Imports libraries, plugins, themes, completions. Prepends existing user `custom/<library>` directories to the path to  allow users to override specific functions. Use `import` to load plugins, themes, handle plugin inter-dependency, etc. 

> __Note__: Adds __only__ existing directories to the path.

### Examples
+ `import plugins/dpaste themes/bobthefish`
+ `import plugins/{cask,brew,django}`

### Extras

+ Also included in this commit is the [`msg`](https://github.com/bucaran/oh-my-fish/blob/import/plugins/msg/msg.fish) plugin. See #285.
msg

### Related & Other Issues

[\#286](https://github.com/bpinto/oh-my-fish/pull/286) [\#284](https://github.com/bpinto/oh-my-fish/pull/284) #285 #96


bpinto:
Let's start the review... :)

bucaran:
Take your time and let me know anything.

functions/_prepend_tree.fish
```diff
@@ -1,72 +1,42 @@
...
-#             -a -and
-#                 Any matches must meet all listed criteria.
-#
-#                     glob1 -a glob2  →  glob1 AND glob2
-#
-#      -d <DESTINATION PATH>
-#          Should appear at the end if used. Specifies the name of the
-#          global path variable to prepend the matched directories to.
-#          If not used, the $fish_function_path is assumed by default.
+#      [glob1 ] -a -and [glob2]
```
bpinto: `[glob1] [-o -or] [glob2] ...`
`[glob1 ] -a -and [glob2]`

could share the same pattern.

`[<glob> [<operator> <glob>..]]` could possibly too.

bucaran: I agree, however the reason for `[-o -or]`, but only `-a -and` is that the `or` operator is optional (used by default if no operator is used), whereas `and` must be specified if you are planning to use it.

functions/import.fish
```diff
@@ -0,0 +1,53 @@
...
+
+    # Prepend plugins and completion paths to fish_function_path and
+    # fish_complete_path respectively. The root paths are determined
+    # via the global $fish_path and $fish_custom.
+    for path in $plugins $completions
+      eval "_prepend_path "$path
+    end
+
+    # Traverse the library tree and prepend directories with fish files.
+    _prepend_tree $fish_path/$library
```
bpinto: Isn't this doing the same thing as 
```
+    set -l plugins     $paths[1..2]" fish_function_path"
+    for path in $plugins $completions
+      eval "_prepend_path "$path
+    end
```

functions/import.fish
```diff
@@ -0,0 +1,53 @@
...
+
+    # Source each library (plugin, theme,...) configuration load file.
+    for path in $root/$library/(basename $library).load
+      [ -e $path ]
+        and eval ". "$path
```
bpinto: Not very fond of `eval`, I'd write this as:

```
for path in $root/$library/(basename $library).load
  if test -e $path
    . $path
  end
end
```

bucaran: Agreed and good catch. Will fix it right away.

functions/import.fish
bucaran: What do you mean by _this_? Why am I seeing a part of the code detached from the function?

oh-my-fish.fish
```diff
@@ -1,74 +1,32 @@
...
+# Initial Configuration
+##
 
 # Set fish_custom to the path where your custom config files
-# and plugins exists, or else we will use the default custom.
-if not set -q fish_custom
-  set -g fish_custom  $fish_path/custom
-end
+# and plugins exist, or use the default custom instead.
+not set -q fish_custom
```
bpinto: The `if` version is more readable.

oh-my-fish.fish
```diff
@@ -1,74 +1,32 @@
...
-end
-
-# Add all defined plugins
-for plugin in $fish_plugins
-  _fish_add_plugin $plugin
-  _fish_add_completion $plugin
-  _fish_source_plugin_load_file $plugin
-end
+# Add functions defined in oh-my-fish/functions to path.
+not contains $fish_path/functions/ $fish_function_path
```
bpinto: The `if` version is more readable.

functions/import.fish
bucaran: Oh wait, you do realize that I need to use `eval` because I am doing brace expansion. That's why we are able to simplify a lot.

oh-my-fish.fish
bucaran: I don't have a strong opinion on this. I prefer not using `if` when one liners are possible, and use `if` for > 1 liners. I can revert it if you really want to.

plugins/rbenv/rbenv.load
```diff
@@ -1,7 +1,11 @@
-set -l rbenv_dir "$RBENV_ROOT"
+set -l rbenv_dir "$ec"
```
bpinto: Another pull request?

plugins/msg/msg.fish
```diff
@@ -0,0 +1,472 @@
+#    __ _  ___ ___ _
```
bpinto: Separate msg plugin in another pr?

functions/_prepend_tree.fish
bpinto: :+1:

plugins/rbenv/rbenv.load
bucaran: :grinning: It was crashing my terminal when I tried loading the plugin so I figured it is loosely related, considering this PR modifies a lot of the foundational elements in the framework.

plugins/msg/msg.fish
bucaran: Let me submit another PR for import first.

functions/import.fish
bpinto: _this_ is the line of code that comment was added.

Thorough example:

Considering my plugin is called: _my_plugin_
```
set -l root $fish_path $fish_custom
set -l paths $root/$library $root/$library/completions
set -l plugins     $paths[1..2]" fish_function_path"

for path in $plugins $completions
  echo "_prepend_path "$path
end
```

Output:

`_prepend_path /Users/bpinto/.oh-my-fish/my_plugin fish_function_path`
`_prepend_path /Users/bpinto/Dropbox/Documents/Bruno/dotfiles/oh-my-fish/my_plugin fish_function_path`


The first `_prepend_path` would be the included in `_prepend_tree $fish_path/$library`. So I don't think we need that `_prepend_path` logic for plugins, only for completions. Or did I miss something? :confused:

oh-my-fish.fish
bpinto: I'd feel more comfortable. I think people are more used to write an `if` clause in here instead of using boolean logic.

plugins/rbenv/rbenv.load
bpinto: I think this will break for other people, I don't have $ec defined on my computer for instance. So, I guess this would not work for me.

oh-my-fish.fish
bucaran: Cool, let's use the `if` version. I hope you let me get away with the one liners in my own plugins though :smiley:

bpinto: I hope I can convince everything is owned by the community. :P

plugins/rbenv/rbenv.load
bucaran: I don't use `rbenv` so I don't mind ignoring this, but I don't understand why it would break for other people. It will only _not_ break for people without `rbenv`, like me.

oh-my-fish.fish
bucaran: Of course you know I mean the plugins I wrote, like `msg`. :stuck_out_tongue_closed_eyes:

functions/import.fish
bpinto: I don't see why this wouldn't work in here. Can you explain it to me?

oh-my-fish.fish
bpinto: :+1:

functions/import.fish
bucaran: My bad, we don't need `eval` for the last loop. We _do_ need it for the one with brace expansion.

`eval "_prepend_path "$path`

bpinto: :+1:

functions/import.fish
bpinto: Having said that, what about `_prepend_tree $fish_path/$library` for `$fish_custom`?

bucaran: Excellent catch.

Useless and inconsequential trivia: actually _not_ in the original `_prepend_tree` version, but later I changed the main loop in _prepend_tree` to

```fish
  for dir in $path $path/**/
  ...
```

effectively including the same directory and making you right about this.

plugins/rbenv/rbenv.load
bpinto: Why would this be executed if you don't have rbenv? I assume you don't have it set on your $fish_plugins.

That line is using my exported $RBENV_ROOT environment variable to define the rbenv dir. If we remove that, then it won't work for my customized install dir.

bucaran: It will crash my terminal if I try to load the plugin when not having `rbenv`.

bpinto: I don't know what happen to the other plugins if you don't have the dependencies installed.
I think this is fine.

bpinto:
I'll do some debugging, it's noticeable slower to start my terminal with this branch.

plugins/rbenv/rbenv.load
bucaran: This was the only one giving trouble in my machine, but you are right, we could improve this later. Will remove in the next `import` PR.

functions/import.fish
bucaran: __However__, we still want to include `$plugins` in the first loop in order to add user custom directories that can override plugin functionality. The fix, is changing:

`set -l plugins     $paths[1..2]" fish_function_path"` __to__ `set -l plugins     $paths[2]" fish_function_path"`.

The first index refers to `$root/$library` which is the actual plugin directory already added by `_prepend_tree`.

bucaran:
Updated PR.

bpinto:
Just tried, still not as fast. I believe the `eval` is causing it, let me try one thing.

bucaran:
I got it.

plugins/rbenv/rbenv.load
```diff
@@ -1,3 +1,8 @@
...
```
bpinto: There is still a bug on this code, don't worry about rbenv for now.

```
.oh-my-fish:prepend_path λ set rbenv_dir abc
.oh-my-fish:prepend_path λ echo $rbenv_dir
abc
.oh-my-fish:prepend_path λ set rbenv_dir "$I_DONT_EXIST"
.oh-my-fish:prepend_path λ echo $rbenv_dir

```

bucaran:
Just pushed new changes. It seems to be loading just as fast as your `master`.

functions/_prepend_tree.fish
```diff
@@ -74,68 +44,48 @@
...
+  # Travese $path and prepend only directories with matches.
+  for dir in "$path" "$path"/**
+    # Use head to retrieve at least the first match.
+    if [ -z (find $dir $glob -maxdepth 1 | head -1) ]
+      continue
+    end
+    if contains -- $argv[1] -p --preview
+      printf "%s\n" $dir
+    else
+      _prepend_path $dir fish_function_path
```
bpinto: https://github.com/bpinto/oh-my-fish/pull/290 alter this.

functions/import.fish
```diff
@@ -0,0 +1,55 @@
...
+  for library in $argv
+    # Prepend plugins, themes and completions, as well as user custom paths
+    # for plugins and themes to $fish_function_path and $fish_complete_path.
+    # The root paths are determined via $fish_path and $fish_custom globals.
+    for path in $fish_custom
+      _prepend_path $path fish_function_path
+    end
+    
+    for path in $fish_path/$library/completions
+      _prepend_path $path fish_function_path
```
bpinto: This should be `fish_complete_path`

functions/import.fish
```diff
@@ -0,0 +1,55 @@
...
+#      functions/_prepend_tree.fish
+#
+# v.0.1.1
+#/
+function import -d "Load libraries, plugins, themes, etc."
+  for library in $argv
+    # Prepend plugins, themes and completions, as well as user custom paths
+    # for plugins and themes to $fish_function_path and $fish_complete_path.
+    # The root paths are determined via $fish_path and $fish_custom globals.
+    for path in $fish_custom
```
bpinto: `$fish_custom/$library`

functions/_prepend_tree.fish
bucaran: So, how do you proceed here? Do you merge your commit and then the import branch? Or do you want me to alter anything?

functions/import.fish
bpinto: This does not allow custom plugins to have subfolders, there is no reason for that.

functions/import.fish
bpinto: With the new `_prepend_path` we would be able to call `_prepend_path $fish_path/$library/completions -d fish_function_path` instead.

functions/_prepend_tree.fish
bpinto: I think it would be better as we can simplify the calls above. Otherwise I'd merge that #290 later.

functions/import.fish
bucaran: Subfolders are handled by `_prepend_tree $fish_path/$library`

bpinto: But if they are custom, they are not under `$fish_path`, but `$fish_custom`.
I think we need two `_prepend_tree` calls, one for `$fish_path` and another for `$fish_custom`.

functions/_prepend_tree.fish
bpinto: Merged #290.

functions/import.fish
bpinto: I think we are talking about different things.

```
# I have a custom boot2docker plugin
.oh-my-fish:master λ echo $fish_plugins
brew boxen boot2docker bundler cliniko dpaste extract ndenv osx rails z rbenv

# My custom folder is not under ~/.oh-my-fish directory.
.oh-my-fish:master λ echo $fish_custom
/Users/bpinto/Dropbox/Documents/Bruno/dotfiles/oh-my-fish

# The custom plugins has a nested folder
.oh-my-fish:master λ ls $fish_custom/plugins/boot2docker/
boot2docker.load clean.fish       nested_folder
```

bucaran: Ah, I see!

functions/import.fish
bucaran: You mean of course replacing both loops for:
`_prepend_path $fish_path/$library -d fish_function_path`
 `_prepend_path $fish_path/$library/completions -d fish_complete_path`

bpinto: :+1:

bucaran: Because of `_prepend_tree` we don't need `_prepend_path $fish_path/$library -d fish_function_path`.

bucaran:
This merges your changes, and updates calls to the new `_prepend_path`.

bucaran:
I added another small improvement during the initial setup.

```fish
# Prepend all user custom paths to the fish path and source load files. 
for custom_file in $fish_custom/**
  _prepend_path $custom_file -d fish_function_path
  switch $custom_file
    case \*.load
      . $custom_file
  end
end
```

In the same loop we are using to source the .load files we can also `_prepend_path` only those that are directories.

bpinto:
I'll to another full review, but if you could squash and rebase, that would be the next step.

functions/_prepend_tree.fish
```diff
@@ -74,68 +44,48 @@
...
 
-  [ -z "$glob" ]
-    and set glob -name \*.fish
-
-  for directory in $source/**/
-    if not [ -z (find $directory $glob -maxdepth $depth | head -1) ]
-      eval $preview # skip when on preview mode
-        and printf "%s " $directory
-        or _prepend_path $directory $destination
+  # Travese $path and prepend only directories with matches.
```
bpinto: Typo traverse.

oh-my-fish.fish
```diff
@@ -1,74 +1,34 @@
...
+# customized via the $fish_path/custom directory. To customize a theme,
+# create a directory under $fish_path/custom/themes with the same name
+# as the theme. Use the same approach for plugins, etc.
+import plugins/$fish_plugins themes/$fish_theme
 
-# Source all files inside custom folder
-for config_file in $fish_custom/*.load
-  . $config_file
+# Prepend all user custom paths to the fish path and source load files. 
+for custom_file in $fish_custom/**
```
bpinto: I don't understand why custom folder is handled differently. I could add a dependency from a custom plugin to another custom plugin.

functions/import.fish
```diff
@@ -0,0 +1,46 @@
...
+  for library in $argv
+
+    # Prepend plugins, themes and completions. Also traverse the library
+    # tree prepending directories with fish code. This should encourage
+    # to create project trees with fish code organized in directories.
+    _prepend_tree $fish_path/$library
+    _prepend_path $fish_path/$library/completions -d fish_complete_path
+
+    # Source each library (plugin, theme,...) configuration load file.
+    for path in $root/$library/(basename $library).load
```
bpinto: Root is undefined.

oh-my-fish.fish
bpinto: By not using the same handler (import function) we are also not filtering the custom plugins using the fish_plugins env var

functions/import.fish
bucaran: Caught.

oh-my-fish.fish
bucaran: Because the path is not $fish_path but $fish_custom, of course.

functions/_prepend_tree.fish
bucaran: Noted.

oh-my-fish.fish
bucaran: I just fixed a bug in `_prepend_tree` where the wildcard was not correctly expanding and not running the for loop at all. fish-shell/issues/683

Also and just for the record, it's good to know to have your wildcards clear:

+ `$fish_custom/**` Traverses both files and directories/subdirectories. Everything. 
+ `$fish_custom/**/` Traverses __only__ directories/subdirectories. (This was another bottleneck :smile:)

functions/import.fish
bucaran: This should simply be `$fish_path` as `import` is not aware of anything else but the fish path. Fixed.

oh-my-fish.fish
bucaran: @bpinto 

### `$fish_path` vs `$fish_custom`
OK. It is handled different _because_ the path we are prepending to `$fish_function_path` is __not__ `$fish_path` but `$fish_custom`.

```fish
for custom_file in $fish_custom/**
  _prepend_path $custom_file -d fish_function_path
  switch $custom_file
    case \*.load
      . $custom_file
  end
end
```
We don't need to call `import` for custom because we are already prepending __everything__ in custom to the path and sourcing all `.load` files by default.

> By not using the same handler (import function) we are also not filtering the custom plugins using the fish_plugins env var.

`$fish_plugins` is just a list with plugin names that exist in `~/$fish_path_plugins`. You can not add custom plugins via `$fish_path`. Those you put in `/custom`. Of course, plugins not under source control inside `/plugins` will always load anyway, but that's not what we are talking about here I presume.

## Wildcards
I just fixed a bug in `_prepend_tree` where the wildcard was not correctly expanding and not running the for loop at all. See here also [fish-shell/issues/683](https://github.com/fish-shell/fish-shell/issues/683)

Also and just for the record, it's good to know to have your wildcards clear:

+ `$fish_custom/**` Traverses both files and directories/subdirectories. Everything. 
+ `$fish_custom/**/` Traverses __only__ directories/subdirectories.

bucaran:
This is it. BTW now `import` prints matched directories by default, but you can skip this redirecting `stdout`. This seems the more unix-like approach I think. Everything should be working now. I will squash into one single nice commit and push again or create another PR.

oh-my-fish.fish
bpinto: > We don't need to call import for custom because we are already prepending everything in custom to the path and sourcing all .load files by default.

This is not the existing behaviour. Custom plugins are just like builtin plugins, they are loaded only if you have them defined on `$fish_plugins`.
There is no reason to not add `$fish_custom` to `import`, we are probably going to do something similar in the future once we add a [vundle like system](https://github.com/bpinto/oh-my-fish/issues/259).

bucaran: Let's recapitulate, there is `$fish_path` (usually `/.oh-my-fish`) and `$fish_custom` (usually `/.oh-my-fish/custom`), these two are _paths_.

There is also a global _list_ of enabled plugins `$fish_plugins`. Plugins (the plugin name) you add to `$fish_plugins` will be handled by `import` during the initial setup. Plugins as well any other user functions (`.load` files too) defined in the `custom` will be added to the path and sourced respectively.

bpinto:
```
#Only plugins added to $fish_plugins are loaded.
for plugin in $fish_plugins
  _fish_add_plugin $plugin
  _fish_add_completion $plugin
  _fish_source_plugin_load_file $plugin
end

function _fish_add_plugin
  set -l plugin $argv[1]
  set -l plugin_path "plugins/$plugin"

# It loads the plugin from ~/.oh-my-fish
  _prepend_path $fish_path/$plugin_path -d fish_function_path
# It loads the plugin from ~/.oh-my-fish/custom
  _prepend_path $fish_custom/$plugin_path -d fish_function_path
end
```

```
# All files with .load inside custom folder are automatically loaded.
for config_file in $fish_custom/*.load
  . $config_file
end
```

bucaran: I see what you mean, but the current implementation does not alter this behavior:

```fish
for custom_file in $fish_custom/**
  _prepend_path $custom_file -d fish_function_path
  switch $custom_file
    case \*.load
      . $custom_file
  end
end

bpinto: We are loading all custom plugin now, if they were defined on `$fish_plugins` or not.

This loop will go through each file and call `_prepend_path` on it. But this function should only be called for directories. So we will be calling this function many times unnecessarily (for each file).

Only `.load` files under `$fish_custom` root need to be searched and loaded. And the rest behaves just like the `import` function.

bucaran: OK, give me a few minutes.

bucaran: Let's continue the conversation over the new PR discussion area. This PR is closed.

PR #289, opened 2015-01-04
remember-last-dir plugin

matklad:
Simple plugin to remember last working directory and 'cd' to it on startup.

There are two main use cases for me: 
* continuing work on some project after reboot
* splitting terminal window to launch tests/build/etc as a side job

I'm absolutely new to fish, so the code may need cleanup.

### DIFF
--- a/plugins/remember-last-dir/remember-last-dir.load
+++ b/plugins/remember-last-dir/remember-last-dir.load
@@ -1,0 +1,5 @@
+function __remember_pwd_changes --on-variable PWD
+        set -U LAST_DIR $PWD
+end
+
+cd $LAST_DIR


bucaran:
If you use [iTerm2](http://iterm2.com/), you can create a custom profile or edit the default one to reuse the previous session's directory when creating new windows / tabs. You would still need a global variable and `cd $LAST_DIR` to remember the last directory across reboots, but other than that, don't you think this would be more suitable for a user defined function instead?

During `oh-my-fish`'s startup, and after plugins, completions and the selected theme is loaded, the following code:

```fish
for config_file in $fish_custom/*.load
  . $config_file
end
```

runs effectively sourcing any `.load` files inside your `oh-my-fish/custom/` directory. This directory is not under `git` control so you can add any personal configuration / globals and/or commands you wish in there.

bpinto:
:+1:

PR #290, opened 2015-01-06
Function _prepend_path allowing multiple paths to be specified at once

bpinto:


bpinto:
@bucaran 
With this, we don't need the `eval` anymore.

bpinto:
This is not to replace the `import` function, just an update to `_prepend_path` to make it more capable and readable.

PR #291, opened 2015-01-07
[feat-import] + import command, ⁑ bad wildcard expansion bugfix in _prepend_tree.

bucaran:
### What's new?

Add new `import` command for plugins, simplifiying intial configuration setup in `oh-my-fish.fish`. Fix incorrect wilcard expansion bug in `_prepend_tree` and other typos in comments.

Let's pretend yesterday didn't happened in this new and immaculate PR. This should be good for merge now, but let me know if there are still any rough corners. 

→ #288

### DIFF
--- a/functions/_prepend_tree.fish
+++ b/functions/_prepend_tree.fish
@@ -1,72 +1,44 @@
-#
 # NAME
-#      _prepend_tree - add a dependency tree to a path
+#      _prepend_tree - add a dependency tree to fish_function_path
 #
 # SYNOPSIS
-#      _prepend_tree [-P --preview] <source path> [<glob>..]
-#                    [-d --destination <destination path>]
+#      _prepend_tree [-p --preview] <path> [<glob>..]
 #
 # DESCRIPTION
-#      Traverse the source path and prepend each directory matching the
-#      glob list to the destination path or to the FISH function path.
-#      If no globs are specified, match any directories with `.fish`
-#      files. Return 1 if the source path is not specified.
+#      Search a path tree and prepend directories with fish files
+#      printing any matches by default. Use a glob list to include
+#      or exclude other file extensions. Use -p --preview to just
+#      print matches withouth modifying the path.
 #
 # OPTIONS
-#      -P --preview
-#          Use to skip adding anything to the path and echo all matched
-#          directories to stdout. Useful for debugging/testing.
+#      [-p --preview]
+#          Do not modify the path. Print directories that match the glob.
 #
-#      <SOURCE PATH>
-#          Required. Specify the path to find glob matches to prepend
-#          to the destination path.
+#      <path>
+#          Required. Specify the path to search for glob patterns.
 #
-#      <glob>...
-#          Glob pattern to match when traversing the source path. If a
-#          directory is found, it is prepended to the destination path.
-#          `.fish` is assumed by default if not globs are specified.
+#      [<glob> [<operator> <glob>..]]
+#          Glob pattern to match when traversing the path path.
 #
-#          It is also possible to use logical OR / AND operators to list
-#          multiple globs. If none is used, OR is assumed by default.
-#          The OR operator
+# OPERATORS
+#      [! -not glob]
+#          Negates the following glob.
 #
-#          The following operators are available:
+#      [<glob> -o -or <glob>..]
+#          Default. Must meet at least one listed criteria.
 #
-#             ! -not
-#                 Negates the following exression.
-#
-#                     ! glob1 glob2   →  NOT glob1 or glob2
-#
-#             -o -or
-#                 Any matches should meet at least one listed criteria.
-#
-#                     glob1 -o glob2  →  glob1 OR glob2
-#
-#             -a -and
-#                 Any matches must meet all listed criteria.
-#
-#                     glob1 -a glob2  →  glob1 AND glob2
-#
-#      -d <DESTINATION PATH>
-#          Should appear at the end if used. Specifies the name of the
-#          global path variable to prepend the matched directories to.
-#          If not used, the $fish_function_path is assumed by default.
+#      [<glob> [-a -and <glob>..]]
+#          Must meet *all* listed criteria.
 #
 # EXAMPLES
-#      _prepend_tree $lib_path
-#          Prepends all directories inside $lib_path containing `.fish`
-#          files to $fish_function_path.
+#      _prepend_tree $path
+#          Match directories in $path containing `.fish` files.
 #
-#      _prepend_tree $lib_path -d PATH
-#          Prepends all directories inside $lib_path containing .fish
-#          files to a global variable named PATH.
+#      _prepend_tree $path \*.fish \*.sh
+#          Match directories in $path with either `.fish` OR `.sh` files.
 #
-#      _prepend_tree $lib_path \*.fish \*.sh
-#          Prepends sub directories with either `.fish` OR `.sh` files.
-#
-#      _prepend_tree $lib_path \*.css -a ! _\*.\* -d PATH
-#          Prepends sub directories that have `.css` extension, but do
-#          not start with `_`.
+#      _prepend_tree $path \*.fish -a ! _\*.\*
+#          Match directories with `.fish` files that do not start with `_`.
 #
 # AUTHORS
 #      Jorge Bucaran <jbucaran@me.com>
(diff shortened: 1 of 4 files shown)


functions/_prepend_tree.fish
```diff
@@ -74,68 +46,52 @@
...
+  # which is safe to use in the loop.
+  set -l subs $path/**/
+  
+  # Traverse $path and $subs prepending only directories with matches.
+  for dir in $path $subs
+    # Use head to retrieve at least the first match.
+    if [ -z (find $dir $glob -maxdepth 1 | head -1) ]
+      continue
+    end
+    printf "%s" $dir
```
bpinto: I'm not sure about this. IMO this should be part of a `--verbose` option.

bpinto:
One question, why not reuse the same PR?

functions/_prepend_tree.fish
bucaran: I thought you didn't like `eval` :grinning: 
Seriously, though, this way is more inline with the way other Unix-like commands work. The default is to print to stdout and it's up to the client to redirect if not needed.

bucaran:
Because I screwed up my branch handling locally. I thought `import` was fine and merged with `msg` and the other rather large plugin I've been working instead of following a more healthy workflow.

bpinto:
> Let's pretend yesterday didn't happened in this new and immaculate PR.
HAhahahaha :+1:

functions/_prepend_tree.fish
bpinto: I feel like `--preview` was useful while creating this function but I don't see any use case for this outside the scope of testing the `_prepend_tree` function itself.

If that assumption is correct, I think this is a good place to use `fish-tank` instead.

bucaran: Mmmm I see, that's actually a good point. Let me change `-p` to `-v --verbose` and push.

bpinto:
It's been great to have this back-and-forth conversation with you! I feel like we are doing a massive improvement to `oh-my-fish`. :+1:

oh-my-fish.fish
```diff
@@ -1,74 +1,34 @@
...
-  end
-
-  if test -e $fish_custom/$load_file_path
-    . $fish_custom/$load_file_path
-  end
-end
-
-function _fish_load_theme
-  _prepend_path $fish_path/themes/$fish_theme -d fish_function_path
-  _prepend_path $fish_custom/themes/$fish_theme -d fish_function_path
```
bpinto: This also allowed custom themes to be loaded from the custom folder.

bucaran:
I've enjoyed this exchange as well very much!

oh-my-fish.fish
bucaran: This behavior still remains.

bpinto: Where?

bpinto: I don't see how this could be working. What is probably happening is that as we are adding all folders inside `$fish_custom` to the path, we are increasing the priority of the `fish_prompt` under custom folder. But even if you don't have this theme set, this will happen, which is not correct.

Also, all files (not only root files) with `.load` under custom folder are automatically loaded, including from plugins that are not set under `$fish_custom`.

bucaran: Yes, that's correct. So, you do want to source all .load files inside custom, but only set plugins defined in $fish_plugins, whether they are in $fish_path or $fish_custom. Is that correct?

bpinto: Yep, that's it. (plugins and themes).

The reason why we automatically load all files under custom folder is so you can add aliases or define environment variables there. I prefer doing it over there than doing it under `config.fish` as I can have several files per project, etc.

And the plugins/themes behave just like the builtin ones, they just live in a different folder, so you can customize it or create your owns without making git dirty.

bucaran: OK, there is a clean solution to this. We just need to let `import` add plugins/themes/completions we load in the initial setup to `$fish_custom` as well:

```fish
import plugins/$fish_plugins themes/$fish_theme
```

And we can get rid of the `for` loop sourcing `.load` files as import will do that too.

bpinto: :+1:

functions/_prepend_tree.fish
```diff
@@ -74,68 +46,52 @@
...
-        or _prepend_path $directory $destination
+  # Null wildcard expansion will break the for loop even if $path is valid.
+  # $subs will become an empty list for directories without sub directories
+  # which is safe to use in the loop.
+  set -l subs $path/**/
+  
+  # Traverse $path and $subs prepending only directories with matches.
+  for dir in $path $subs
+    # Use head to retrieve at least the first match.
+    if [ -z (find $dir $glob -maxdepth 1 | head -1) ]
```
bpinto: Could you explain to me what is this for? Is this similar to `test -d (File is directory)`?

bucaran: Correct.

bucaran:
Done.

+ All `.load` files inside custom are sourced as usual.
+ Only set plugins set in `$fish_plugins` as imported. Works whether they are in `$fish_path` or `$fish_custom` as expected. The same for plugins.
+ `$fish_function_path` is not polluted.

Note: Make sure to restart the terminal after modifying your `.config` file as it seems this file is only loaded once when the Fishshell startup. Reloading `oh-my-fish.fish` does not reload

bucaran:
@bpinto

oh-my-fish.fish
bucaran: /play trombone

functions/_prepend_tree.fish
```diff
@@ -1,72 +1,44 @@
...
+#      _prepend_tree [-p --preview] <path> [<glob>..]
 #
 # DESCRIPTION
-#      Traverse the source path and prepend each directory matching the
-#      glob list to the destination path or to the FISH function path.
-#      If no globs are specified, match any directories with `.fish`
-#      files. Return 1 if the source path is not specified.
+#      Search a path tree and prepend directories with fish files
+#      printing any matches by default. Use a glob list to include
+#      or exclude other file extensions. Use -p --preview to just
```
bpinto: I thought you were doing `--verbose` instead.

functions/_prepend_tree.fish
```diff
@@ -74,68 +46,64 @@
...
-  [ -z "$glob" ]
-    and set glob -name \*.fish
+  # Null wildcard expansion will break the for loop even if $path is valid.
+  # $subs will become an empty list for directories without sub directories
+  # which is safe to use in the loop.
+  set -l subs $path/**/
+
+  # Traverse $path and $subs prepending only directories with matches.
+  for dir in $path $subs
+    # Use head to retrieve at least one match. Ignore errors for non
```
bpinto: How can it be possible to have non-existing directories?

functions/import.fish
```diff
@@ -0,0 +1,47 @@
...
+#      functions/_prepend_path.fish
+#      functions/_prepend_tree.fish
+#
+# v.0.1.0
+#/
+function import -d "Load libraries, plugins, themes, etc."
+  for library in $argv
+    # Prepend plugins, themes and completions, traversing library
+    # trees and prepending directories with fish code.
+    _prepend_tree $fish_path/$library
```
bpinto: I am missing something here.

`$fish_path -> ~/.oh-my-fish/`
`$library -> dpaste`
Where is the `plugin` folder?

functions/import.fish
```diff
@@ -0,0 +1,47 @@
...
+    _prepend_tree $fish_custom/$library
+    _prepend_path $fish_path/$library/completions -d fish_complete_path
+
+    # Set path to load files.
+    set -l path $library/(basename $library).load
```
bpinto: `load_files_path`?

functions/_prepend_tree.fish
bpinto: Nevermind, I figured it out. When we send a non-existing directory as argument to this function.

functions/import.fish
bucaran: Sorry, you are right, fixed.

functions/_prepend_tree.fish
bucaran: Remember we are calling `_prepend_tree` with `$fish_custom`. This will print a silly `not found` error message when `find` gets passed a directory that does not exist.  _Improved the comment a bit in the fix._

functions/_prepend_tree.fish
bucaran: Thanks for pointing that one out.

functions/import.fish
bucaran: When loading plugins or themes, you need to specify the `plugins/` or `themes/` base.

```
 import plugins/<plugin>
 import plugins/{dpaste,cask} themes/bobthefish
```

functions/import.fish
bucaran: Same block, but doing different things so the comments are different. Well, I don't mind shifting these two lines and pushing another commit, but this way is just a bit more readable to me...

bpinto:
Everything I tested is working now. :+1:

functions/import.fish
bpinto: Code style, I won't bug you with this.

oh-my-fish.fish
bucaran: > And the plugins/themes behave just like the builtin ones, they just live in a different folder, so you can customize it or create your owns without making git dirty.
:+1:

bpinto:
Tell me when this branch is ready and I'll merge it (FINALLY!) :+1:

functions/import.fish
bucaran: Not at all. :smile:

bucaran:
Everything is working fine here. Custom plugins and other plugins ok as well, themes are changing accordingly. I think this is it. :+1: :+1:

bpinto:
Awesome! Merged!

bucaran:
Next stop: _plugins_.

bpinto:
Any sneak peek? :P

bucaran:
It's almost ready, so I guess I can say what it is about.

I know there is `hub` written in Go and `git-flow`, but I still wanted something simpler for everyday stuff and without all the dirty details. So I set out to write a `git` + `github` workflow friendly wrappers.

```
#    ____
#   / ___\
#  / /_/  >
#  \___  /
# /_____/ v0.1.0
#
# NAME
#      g - git + github workflow spells
#
# SYNOPSIS
#      g [d download] [D delete] [e explore] [f fork] [g get] [gi] [h help]
#        [i io] [k like] [l list] [n new] [o open] [s sync] [t lift] [v ver]
```

bpinto:
Interesting, I might use it! :+1:

bucaran:
Well, I will try to use it for simple stuff and see if it can evolve into something more robust without losing the simplicity model. The API design is also worth mentioning, the main command `g` acts as a router for the CLI layer that describes each of the available _spells_ individually. These spells are concerned with parsing their options, printing error/success status to stdout, etc. There is also an API layer, that bridges the CLI layer with the actual `curl` requests to the github API and most involved `git` flows.

bucaran:
I've gone through `msg` again and my tests are running well so I am going to push now and continue working on `g`. The code has not changed much from the last time I PR `msg`. I just added more comments and started using `if` consistently instead of intercalating `and/or` blocks every now and then.

bpinto:
I like the utilities library, I'll review the `msg` plugins probably later tonight.

bucaran:
Thank you. Let me know anything.

PR #292, opened 2015-01-07
Add msg plugin. The technicolor message printer.

bucaran:
#### Features
 + Print messages with style and color easily. Supported styles below:
```
     _text_                 Bold
     __text__               Underline
     ___text___             Bold and Underline
     `$variable`            Apply @<styles> to $variables
     /directory/            Directories
     [url]                  Links
     \n                     Line Break
     \t                     Tab Space

     @<fg:bg> fg=bg=RGB|RRGGBB|name|random
       RGB value can be any three to six hex digit or color name.
       e.g, @0fb, @tomato, @random, @error, @success/ok
```
 + Recognizes 256 different color names and special `@random` `@light` `@dark` `@error` and `@success` labels.
 + Included also is `msg.ask` to print a message with `msg` and wait for y/n input. Returns true on y\*.

### Example
```fish
function msg_test
  set -l dot @random:random "   "
  for 1 in (seq 6)
    set dot $dot $dot
  end
  msg -sn @black:yellow "    Loading ...  " $dot
  msg @success ✔ OK
  msg _msg_ the @yellow:red technicolor __printer__!\
    \n A _brave_ new __line__.\
    \n [http://github.com/oh-my-fish] @red → \\[url]\
    \n /dir/ect/to/ri/es/ @red → \\/directories/
end
```
41c69d96-942b-11e4-853e-c5be2d8be68d


bpinto:
Is this ready?

bucaran:
@bpinto ping

plugins/msg/msg.fish
```diff
@@ -0,0 +1,523 @@
...
+set --global msg_color_fg   FFFFFF
+set --global msg_color_bg   normal
+set --global msg_style_url  00FF00 $msg_color_bg -u
+set --global msg_style_dir  FFA500 $msg_color_bg -u
+set --global msg_color_err  FF0000
+set --global msg_color_ok   00FA9A
+
+# Set style, fg/bg colors and reset. Modifies its parent scope of `msg`.
+# @params [<fg>] [<bg>] [<style>]
+function --no-scope-shadowing msg.util.set.color
```
bpinto: I think each of these functions should be on their own file to make use of lazy loading and also it's a more fish way.

plugins/msg/msg.fish
```diff
@@ -0,0 +1,523 @@
...
+end
+
+function msg -d "Technicolor printer."
+  # Default " " whitespace between strings, skip with -s.
+  set -l ws " "
+  # Default \n newline after message, skip with -n.
+  set -l ln \n
+
+  switch (count $argv)
+    # No hard feelings, but don't print nothing.
```
bpinto: Did you mean 'don't print anything.'?

plugins/msg/msg.fish
```diff
@@ -0,0 +1,523 @@
...
+    case 0
+      return 0
+    case \*
+      switch $argv[1]
```
bpinto: This is a weird `switch` usage as there is only one `case`.
I think you could remove this switch and code would become smaller.

plugins/msg/msg.fish
```diff
@@ -0,0 +1,523 @@
...
+              set ws ""
+            end
+            # Use -n to not add a newline at the end of the message.
+            if msg.util.str.has "n" $argv[1]
+              set ln ""
+            end
+            # We are done with options, get rid of first item.
+            set argv $argv[2..-1]
+          else
+            return 0
```
bpinto: Isn't this the case of having no options (count argv == 1)? Shouldn't it print the message?

plugins/msg/msg.fish
```diff
@@ -0,0 +1,523 @@
...
+  # and use the reset. set_color normal makes sure to reset both colors and
+  # bold / underline styles.
+  set -l reset (set_color normal)(set_color $msg_color_fg)
+
+  # Foreground and background color carried from previous @style token.
+  # These variables are set and reset via msg.util.set.color
+  set -l fg $msg_color_fg
+  set -l bg $msg_color_bg
+
+  for token in $argv
```
bpinto: I like this!

plugins/msg/msg.fish
```diff
@@ -0,0 +1,523 @@
...
+
+  if [ (count $argv) -gt 2 ]
+    set_color $argv[3]
+  end
+
+  set bg $msg_color_bg
+  set fg $msg_color_fg
+end
+
+function msg -d "Technicolor printer."
```
bpinto: It's quite hard to read this code, it's too long. I'd split into many small descriptive functions.

bpinto:
`msg.test` is gone?

plugins/msg/msg.fish
bpinto: I'm missing something, I thought this `return 0` would break the function and print nothing.

So I thought this wouldn't work:
```
.oh-my-fish:bucaran-plugin-msg λ msg a
a
```

plugins/msg/msg.fish
bucaran: Thanks!

plugins/msg/msg.fish
bucaran: The reason I nest switch statements is so that I can use wildcard expansion. This is actually the simplest way to do this in Fish (unless you give up wildcard expansion, that is).

plugins/msg/msg.fish
bucaran: I mean, since `count < 1` so don't try to print what we don't have. Don't print anything would be more like _skip_ printing, but we can't even do that here since we've got nothing to print! :laughing:

plugins/msg/msg.fish
bucaran: I like this. Will do.

plugins/msg/msg.fish
bucaran: Remember we are inside `case -s\* -n\*`. So, this is handling the absurd case of  `msg -option`.

bucaran: If you want to print `-s` or `-n` you should escape them, e.g., `msg \-s`. However, I forgot to add that option when escaping the other style separators. Already fixed. Will push with next PR.

plugins/msg/msg.fish
bucaran: I am afraid this portion exists in order to reduce the amount of repetition when setting / resetting color and styles. This was added after I had a working version. I could put this in a separate file too, but let me check if that affects the `--no-scope-shadowing`.

bucaran:
Pushed.

Try `msg.test` here:

```fish
function msg.test
  set -l dot @random:random "   "
  for 1 in (seq 6)
    set dot $dot $dot
  end
  msg -sn @black:yellow "    Loading ...  " $dot
  msg @success ✔ OK
  msg _msg_ the @yellow:red technicolor __printer__!\
    \n A _brave_ new __line__.\
    \n [http://github.com/oh-my-fish] @red → \\[url]\
    \n /dir/ect/to/ri/es/ @red → \\/directories/
end
```

plugins/msg/msg.fish
bpinto: You are right, I forgot it was inside the switch.

plugins/msg/msg.fish
```diff
@@ -0,0 +1,164 @@
...
+  set -l ws " "
+  # Default \n newline after message, skip with -n.
+  set -l ln \n
+
+  switch (count $argv)
+    case 0
+      # Nothing to print here.
+      return 0
+    case \*
+      switch $argv[1]
```
bpinto: I thought this would be same as

What if `msg.util.str.has` was added the option to use regex (it uses grep)?

```
if msg.util.str.has "-(s|n){1,2}" $argv[1]
end
```

It might not be an improvement in the end.

bpinto:
I have nothing else to add, I had tested it before and it was working, tell me if it's ready and I'll do another test and merge.

Very well done plugin. :+1:

bucaran:
Thanks! Yes, we are good. I do have plans to add one or another thing in the future. Play more with colors maybe, but v.0.1.0 is ready. :ok:

plugins/msg/msg.fish
bucaran: Hey let me check this.

bucaran: I think this would be nice to implement. I will add it to `0.1.2`. Right now I would like to work on other plugins too.

bpinto: :+1:

bucaran:
:+1:

PR #296, opened 2015-01-10
Simplify remove_from_array function to use builtin contains command.

bucaran:
Better off using the builtin contains, me thinks.

```fish
# Remove item from list. List must be the name of a global variable.
# @params <item> <list>
function remove_from_array
  set -l item $argv[1]
  if set -l index (contains -i -- $item $$argv[2])
    set -e $argv[2][$index]
  end
end
```

### DIFF
--- a/test/test_oh-my-fish.fish
+++ b/test/test_oh-my-fish.fish
@@ -57,14 +57,12 @@ function load_oh_my_fish
   . $fish_path/oh-my-fish.fish
 end
 
+# Remove item from list. List must be the name of a global variable.
+# @params <item> <list>
 function remove_from_array
-  set -l element $argv[1]
-
-  for i in (seq (count $$argv[2]))
-    if test $$argv[2][$i] = $element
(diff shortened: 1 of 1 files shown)


bpinto:
Much better indeed! I didn't know about this flag, I thought I knew a way to do this but couldn't remember and couldn't find any example, thanks!

bucaran:
Fish FTW! :+1:

bpinto:
I'll move this method to a array.remove function as sso

bpinto:
I was thinking about `array.delete "$fish_path/themes/l" --from fish_function_path` instead of `    array.delete "$fish_path/themes/l" fish_function_path`, what do you think?

bucaran:
Just pushed a new version. It's pretty cool. A few comments, AFIK the correct terminology in fish to refer to arrays is _lists_, so I renamed it to `list.erase`.  We are using `set -e --erase` anyway.

PR #294, opened 2015-01-09
Fox theme...

lvicentesanchez:
cloned from oh-my-zsh.

PR #300, opened 2015-01-12
Improve install.fish

bucaran:
This is the first in a series of PRs to improve OMF installation, overhaul new fish-spec test framework, add automated test support, etc.

### `improve-install`

Add support for:

  + Non-interactive terminals, e.g., skip launching fish when
    running tests automated, etc.
  + Non-git-based installations. If git is not found, download
    via curl HTTP GET download request.
  + Installing forks of the framework.

Also:
  + Improve documentation.
  + Fix bug in un escaped backslash causing incorrect syntax
    coloring after the ASCII fish logo.

### DIFF
--- a/tools/install.fish
+++ b/tools/install.fish
@@ -1,68 +1,122 @@
 #!/usr/bin/env fish
+#
+# Oh-my-fish is a user-friendly framework for managing your fish-shell
+# configuration. It includes optional plugins (brew, git, rails, python,
+# node, etc) and themes.
+#
+# Get the party started with:
+#
+#   curl -L https://github.com/bpinto/oh-my-fish/raw/master/tools/install.fish \
+#   | fish
+#
+# Oh-My-Fish currently supports
+#
+#   - Fish 2.0+ [fishshell.com] [github.com/fish-shell/fish-shell]
+#
+# The installation process is split up in the following steps:
+#
+#   1. Resolve source repository to default or get from environment variable.
+#      This allows installers curl -L ..install.fish | FORK=<fork> fish in
+#      order to test forks of the framework.
+#
+#   2. Check for already installed copies of Oh-My-Fish. Do not reinstall.
+#
+#   3. Attempt to pull repository via git clone, if git is not available,
+#      curl a HTTP GET request directly from github.
+#
+#   4. Backup original configuration file if there is one and copy new
+#      configuration file from templates/.
+#
+#   5. Print fish logo.
+#
+#   6. Start prompt if the terminal is interactive
+#/
 
-# Takes color as first argument, and text to print as other arguments.
-function colored
-    set_color $argv[1]
-    set -e argv[1]
-    echo $argv
-    set_color normal
+set fish_path    ~/.oh-my-fish
+set config_path  ~/.config/fish/
+
+# Log a color message.
+#   log <color> <string>...
+function log
+  echo -e (set_color $argv[1])$argv[2..-1](set_color normal)
 end
 
-if test -d ~/.oh-my-fish
-    colored yellow -n You already have Oh My Fish installed.
-    echo " You'll need to remove ~/.oh-my-fish if you want to install"
-    exit
+log white "Installing Oh My Fish..."
+
+# Allow installers to specify the source repository.
+if not set -q FORK
+  set FORK bpinto
 end
 
-colored blue Cloning Oh My Fish...
-type git >/dev/null
-and git clone https://github.com/bpinto/oh-my-fish.git ~/.oh-my-fish
-or begin
-    echo git not installed
-    exit
+# Abort installation if oh-my-fish is already installed.
+if test -d $fish_path
+  log yellow "You already have Oh My Fish installed."
+  log white  "Remove $fish_path if you want to reinstall it."
+  exit 1
 end
 
-colored blue Looking for an existing fish config...
-if test -f ~/.config/fish/config.fish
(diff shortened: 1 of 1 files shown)


tools/install.fish
```diff
@@ -1,68 +1,128 @@
...
+  log white "Installing Oh My Fish..."
 
-colored blue Looking for an existing fish config...
-if test -f ~/.config/fish/config.fish
-    colored yellow -n "Found ~/.config/fish/config.fish."
-    colored green " Backing up to ~/.config/fish/config.orig"
-    mv ~/.config/fish/config.{fish,orig}
-end
+  # Allow installers to specify source repository.
+  if not set -q FORK
```
bpinto: I don't think this is necessary. Usually when you want to contribute to a project, you will fork on github and then just add a remote to git.

This won't help people contributing back to the project, as they won't have an upstream remote to keep their branches up to date.

tools/install.fish
```diff
@@ -1,68 +1,128 @@
...
 
-colored blue Cloning Oh My Fish...
-type git >/dev/null
-and git clone https://github.com/bpinto/oh-my-fish.git ~/.oh-my-fish
-or begin
-    echo git not installed
-    exit
-end
+# Install and setup Oh My Fish. Wrapping the script in a function allow us
+# to defer the installation after the script is completely downloaded.
```
bpinto: Have you had an error with the install before wrapping in a function?

tools/install.fish
bucaran: Actually, I only added this because I needed to test with downloading my fork. Obviously your fork does not have this yet, so the installation would not run and the travis build would fail. I don't really care about having this or not, but I had to use this in order to make travis comply and it seemed like a no-brainer feature to have. Also, it was pretty cool to learn.

tools/install.fish
bucaran: No. But wrapping it in a function is nicer and sort of helps organizing things better because there is also a `log` function. The thing about the script getting downloaded before running will probably never be an issue, but it's also good to know that that is how it works. This is generally speaking more good than bad, because it prevents our output overlapping with curl's. We can't run a partially downloaded script either. Again, may not be a problem, but does not harm and it wraps the process in a descriptive function `install`.

bpinto: Can you show me an example of this in action? I don't understand how the function would prevent this from happening and vice-versa.

Having a function called `install` is not making the code easier to read or anything else, as the file is already called `install`. It's only being redundant to me as it's the same code with or without the function.

tools/install.fish
bpinto: If we need this to make it work on travis, than we should keep it.

Simplicity > Flexibility imo.

bucaran: Hehe, honestly, it was difficult finding out how to do this so I said that bit about flexibility. 

> Simplicity > Flexibility

100% agree.

tools/install.fish
bucaran: I guess you are right about that. Since this is an issue we will probably never run into, I guess we can forget about it. I learned about this when working with [Meteor JS](https://www.meteor.com/). I have never experienced this so I should not really be using it as a good argument!

tools/install.fish
bpinto: :+1:

bucaran:
Squashed and ready.

tools/install.fish
```diff
@@ -1,68 +1,122 @@
...
-colored blue "Using the Oh My Fish template file and adding it to ~/.config/fish/config.fish"
-cp ~/.oh-my-fish/templates/config.fish ~/.config/fish/config.fish
+# Either git clone or curl GET repository.
+log blue "Cloning Oh My Fish from remote repository..."
+if type git >/dev/null
+  git clone "https://github.com/$FORK/oh-my-fish.git" $fish_path
+else
+  log yellow "Install git to pull Oh-My-Fish updates"
+  log white "Downloading remote zip from Github..."
+  set
```
bpinto: What is this?

bucaran: This downloads directly from the repo via curl if user has no git installed. Some users may just want to enjoy the plugins and have nothing to do with git.

bpinto: But this is an loose set call.

bucaran: Yes, indeed. Squashed.

bucaran:
Squashed and ready :+1:

bpinto:
Thanks for this improvement!

P.S.: I thought this was rebased, next time if you could rebase, that would be great.

bpinto:
As a side note, a better way to add commit to git is to have small <80 characters commit message and use the commit description (break line and add an empty line after the message) to add more information.

bucaran:
:+1:

bucaran:
> P.S.: I thought this was rebased, next time if you could rebase, that would be great.

But it was rebased. I squashed all the commits after the original PR into the first one with the long message. I just looked at your commit history and it seems to be correct to what I expect. What went wrong?

bpinto:
The upstream had a merge from PR #294.

bucaran:
Mmm.

I could've made a mistake of course, but just for the record here is what I did each time I had a commit to squash.

`rebase -i HEAD~2`

There I would leave everything intact, and only set the last commit (the one to squash) to `s`.

Did I miss anything?

bpinto:
I usually do the following:

```
git fetch upstream
git rebase -i upstream/master
```

PR #301, opened 2015-01-13
Theme edan update

amio:
Add user & host to left prompt.


themes/edan/fish_prompt.fish
```diff
@@ -53,3 +64,21 @@ function fish_prompt
   echo -n -s ' » ' $normal
 
 end
+
+# Display the compressed current working path on the right
+# If the previous command returned any kind of error code, display that too
+
+function fish_right_prompt
```
bpinto: This is supposed to be on another file, as it used to be.

themes/edan/fish_prompt.fish
```diff
@@ -33,6 +42,8 @@ function fish_prompt
       echo -n -s (set_color -b cyan black) '[' (basename "$VIRTUAL_ENV") ']' $normal ' '
   end
 
+  _user_host; echo -n ': '
```
bpinto: It would be nice to add an environment variable to enable this as people might not want this enabled.

themes/edan/fish_prompt.fish
amio: I was thingking fish seems load these files every time after excuting a command，so one file may reduce some loading time。But if it's a best practice，no problem of course~

themes/edan/fish_prompt.fish
amio: Sure that will be better, thanks for your advice.

themes/edan/fish_prompt.fish
bucaran: The load would be not easily observed so it's overhead IMO, however @amio's clever theory would be worth exploring directly in the fish source code. I don't know for sure if fish will actually load these files every time. Once a function is loaded, it stays in memory during the shell session unless they are explicitly erased with `function -e \`func_name\``.

bpinto: AFAIK it stays in memory.

It will add a little bit of overhead, but I think it will be unnoticeable. But this, for instance, allows people to have a custom `right_fish_prompt` in their repo and overwrite the default one.

amio:
ping @bpinto 
I've add two function `edan-remote` and `edan-local` to switch the user&host info, by default it will works in local mode as previous( no user & host info).

themes/edan/fish_prompt.fish
```diff
@@ -1,9 +1,17 @@
-# name: iden
+# fish theme: edan
 # Display the following bits on the left:
 # * Virtualenv name (if applicable, see https://github.com/adambrenecki/virtualfish)
 # * Current directory name
 # * Git branch and dirty state (if inside a git repo)
 
+function edan-remote
```
bpinto: As these functions are to be used outside of this fish_prompt function, I'd move them to separate files.

bpinto:
Nice change.

themes/edan/fish_prompt.fish
amio: Is it any `.fish` file whithin `theme/edan` folder will be lodaded? Or should I name it like `functions.fish`?

bucaran: Name your files after your functions, e.g., `my_function.fish`.

amio: I have tried `edan_function.fish` or `my_function.fish` but seems neither working.

bucaran: > Is it any .fish file whithin theme/edan folder will be lodaded? Or should I name it like functions.fish?

Yes. It will be loaded during initial setup by `functions/import.fish`. Try naming your files exactly as your two functions, i.e., `edan-remote.fish` and `edan-local.fish` and make sure they exist in your theme directory `themes/edan`. Let me know if _that_ does not work.

amio: That *works* :)

themes/edan/edan-local.fish
```diff
@@ -0,0 +1,5 @@
+# Switch environment to "local" for edan.
+
+function edan-local
+  set -eU EDAN_HOST_TYPE
```
bpinto: The `-U` is not necessary.

themes/edan/README.md
```diff
@@ -1,14 +1,15 @@
-## edan
+## Theme Edan
 
 Inspired by idan, a functional, uncluttered fish theme with usability perks for git users and python developers.
 
-![Screenshot of edan theme](https://cloud.githubusercontent.com/assets/215282/2869575/7bd3f828-d278-11e3-91ce-7ff84a5ea1b0.png)
+![Screenshot of edan theme](https://raw.githubusercontent.com/amio/oh-my-fish/master/themes/edan/Screenshot.png)
```
bpinto: Instead of adding the screenshot to git, we instead upload the screenshot to the PR and use the link on the README so our git repo is not unnecessarily big.

P.S.: The commit that added the screenshot needs to be removed from git log.

themes/edan/README.md
```diff
@@ -1,14 +1,15 @@
...
 
 
 #### Characteristics
 
 ##### Left Prompt
 
+* User & host (hidden by default, excute `edan-remote` to show, `edan-local` to hide)
```
bpinto: Typo: excute.

amio: Ooops

themes/edan/edan-local.fish
amio: As fish document say,
> If a variable is not explicitly set to be either universal, global or local and has never before been defined, the variable will be local to the currently executing function.

Well I hope this variable will be kept even after reboot, for using on a remote host, so I add the `-U` specifically.

themes/edan/README.md
amio: screenshot

themes/edan/edan-local.fish
bpinto: I don't know what you are trying to achive.
>  -e or --erase causes the specified shell variable to be erased

You are going to erase the variable, you don't need to
> -U or --universal causes the specified shell variable to be given a universal scope.

amio: Ahh, thanks. I thought set it with `-U`, then should erase it with `-U` too :P

amio:
I have made another pull request with clean commit history, and remove the screenshot.

PR #302, opened 2015-01-13
Add travis support via .travis.yml and bash bootstrap script.

bucaran:
The bootstrap script is currently installing my fork during the remote built test, again, this is because my build is the one with the most recent changes and to show it works.

Tests are passing! Checkout my master build [here](https://travis-ci.org/bucaran/oh-my-fish/builds/46844547). Notice this is build \#73 which is using my master branch, the most recent one \#74 is actually __failing__, but that's because it's pulling from my `travis` branch (this PR's branch), which does not have the `fish-spec` changes. You can ignore that ATM. 

(1) Let's merge first `travis`, and then (2) review the `fish-spec` branch. After that, you can (3) merge `fish-spec` and (4) add another commit yourself removing `FORK=bucaran` from `script/boostrap.sh`. Finally (5) I will pull from upstream master into my master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+language: c
+
+env:
+  - FISH_PPA=nightly-master
+
+before_install:
+  - script/bootstrap.sh
+
+script: fish ./spec/spec.fish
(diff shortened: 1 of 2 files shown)


bpinto:
Thanks for this!

PR #303, opened 2015-01-13
❯❯❯ Complete overhaul of brand new Oh-My-Fish `fish-spec` test framework.

bucaran:
+ Add documentation and examples.
+ Add new functions to interact with the framework:
  + spec.eval - eval a function by name and echo its description field
  + spec.functions - prints functions currently in scope that match the key glob
  + spec.log - log a message / test result
  + spec.run - run suite of tests
+ Rename list.erase to erase.

There are some cool new things. I believe I improved the previous `spec.suite`, `spec.log.ok` mess and we have a more consistent framework now ready to start growing.

```fish
import plugins/fish-spec

function describe_my_suite -d "Test suite message..."
  function before_all
  end
  function after_all
  end
  function test_1 -d "It should work!"
  end

spec.run $argv
```


bucaran:
`spec.run` is taking a -v --verbose flag to output the full test descriptions. It's off by default. 
There some output by default though. 

I chose ❯ for the success visual cue because as you had pointed out in an earlier review, some tests may specify multiple success conditions, invoking `expect` several times per test case. In this case, spec.log will be run once per each `expect` call thus ❯ characters will stack and that's pretty cool.

plugins/fish-spec/erase.fish
```diff
@@ -13,10 +13,9 @@
 #   While items are basically any valid sequence of symbols, lists refer
 #   to any global variable or local variable in the scope of the calling
 #   function by name.
-#
 #/
-function -S list.erase
-  # List to erase from is last item by default.
+function -S erase
```
bpinto: I think it should be scoped inside `list`. We can build an auxiliary library like that. As we don't have objects, `list` is like a class and `erase` is a class method.

plugins/fish-spec/expect.fish
```diff
@@ -1,34 +1,52 @@
...
+#
+# DESCRIPTION
+#   Asserts a list of expected values match an actual value.
+#
+# EXAMPLE
+#   import plugins/fish-spec
+#
+#   function describe_echo
+#     function it_echoes_a_value
+#       set -l value (echo "Oh My Fish!")
```
bpinto: `set -l value 'Oh My Fish!'`

plugins/fish-spec/erase.fish
bucaran: This way looked more cool, but sure. 

BTW remember I told you about a list manipulation library I was working on? I would like to push that into the framework too as a library. It would be useful if it could be used across the framework.

My library does not have this erase function though. So in the future, after the list library becomes a part of the framework, we can edit update the `fish-spec` to use that library instead.

Another thing, I am using the extension `ls` (stands for list as you know) to scope the methods. `ls` is recognized by the Fish lexer in my editor so it's really nice to the eye.

```fish
ls.erase
```

What do you think?

plugins/fish-spec/expect.fish
bucaran: Not the best example.

plugins/fish-spec/erase.fish
bpinto: I think most people are shell users, and `ls` already have a _different_ meaning.
`list` is much more meaningful.

bucaran: OK. What is that meaning by the way? `ls` means list AFAIK.

plugins/fish-spec/spec.log.fish
```diff
@@ -0,0 +1,59 @@
...
+#   spec.log -m | --msg/--message
+#            -o | --ok
+#            -f | --fail
+#            [<color> <string>...]
+# OPTIONS
+#   -m --msg --message
+#     Output the message according to the <color> <string>... list.
+#
+#   -o --ok
+#     Echo a ❯ character. Optionally, print a list of <color> <string>...
```
bpinto: I think a `.` character looks better, don't you?

bpinto: On the verbose mode, I'd replace `.` with the `√` sign.

plugins/fish-spec/spec.log.fish
```diff
@@ -0,0 +1,59 @@
...
+#     description and actual value of the failed assertion. For example:
+#
+#       spec.log --fail red "Expected" blue "$value"
+#
+# NOTES
+#   This function does not modify the $status variable.
+#
+#   Some tests may specify multiple success conditions, invoking `expect`
+#   several times per test case. In this case, spec.log will be run once
+#   each per `expect` call.
```
bpinto: I think it would be better to only long once pert test, but this is something I can work on once we merge this PR.

plugins/fish-spec/spec.log.fish
bucaran: Well, it's really a matter of taste I guess, but regardless of which one is better, do notice that a `.` will be _harder_ to see on the screen than a ❯, unless the test has multiple `expect` cases, in which is fine.

bucaran: Let's leave it up to users of the framework to configure then.

plugins/fish-spec/spec.log.fish
bucaran: Actually, it's _really_ useful to see which of your expect cases failed in the test. 

```❯ ❯✘ ❯```

File: plugins/fish-spec/erase.fish (hunk 1 above)
Comment (bpinto) [inline plugins/fish-spec/erase.fish]: `ls` is a builtin library. It would become confusing.

File: plugins/fish-spec/spec.log.fish (hunk 3 above)
Comment (bpinto) [inline plugins/fish-spec/spec.log.fish]: When you run tests every few seconds, you don't need to see the `.`, that's why it's better. You just need to quickly see that something bad happened.

I don't think we need to have this as a framework configuration, it's something people are already used to, simplicity is better for the moment.

File: plugins/fish-spec/erase.fish (hunk 1 above)
Comment (bucaran) [inline plugins/fish-spec/erase.fish]: Sure. But I think it's a neat trick to play with Fish lexer, you have to agree that it looks cool. Will fix.

File: plugins/fish-spec/spec.run.fish [hunk 5]
```diff
@@ -0,0 +1,62 @@
...
+  if contains -- $argv[1] -v --verbose
+    set output ""
+  end
+  # Run 1 or more topmost describe blocks causing fish to load inmediately
+  # scoped functions, i.e, those inside any invoked describe_ blocks.
+  spec.eval describe_ --depth 0 $output
+  spec.eval before_all $output
+  for test in (spec.functions "it_")
+    spec.eval before_each $output
+    # The test run does not stop when a test fails, but make sure to
```
bpinto: Future change will be to stop the test run after something failed.

plugins/fish-spec/spec.log.fish
bucaran: OK. Let's `.`, but verbose mode should be nicer. Verbose is off by default.

plugins/fish-spec/test/erase.spec.fish
```diff
@@ -0,0 +1,56 @@
...
+    set -g odds_until_10 1 3 5 7 9
+  end
+
+  function it_erases_one_element -d "It should erase an element from a list."
+    erase 1 --from nums_until_10
+    expect $nums_until_10 --to-not-contain 1
+  end
+
+  function it_erases_one_element_with_from_syntax \
+    -d "it_erases_one_element_with_from_syntax"
```
bpinto: These descriptions are copy and paste from the function name.

plugins/fish-spec/spec.run.fish
bucaran: I don't oppose this, but if you have a good argument against it, I would like to know about it. 
What's your experience with other test frameworks?

plugins/fish-spec/spec.log.fish
bpinto: The error message should say which expectation failed, that should be enough.

Before we start having profiles to output in different ways (we might never need to build something this fancy), I think we can have `--quiet`, `--verbose` and nothing.

`--quiet`: `.....✘...`
normal: `..... Expected x actual y ...`
`--verbose`: `test description √ \n test description √ \n Expected x actual y ✘`

plugins/fish-spec/test/erase.spec.fish
bucaran: Ah, yes, I forgot to fix those. Noted.

plugins/fish-spec/spec.log.fish
bpinto: Or maybe we don't need a quiet, and just output the error message in the end of the suite run, that would be amazing.

bpinto: But anyway, I'll work on this after we get this merged.

bucaran: We definitely need to start writing tests for more stuff and see what works or not, before doing a lot of forecasting.

plugins/fish-spec/erase.fish
bpinto: hahaha it does, but we won't be able to do this for everything. e.g. `string.contains`.

bucaran:
```fish
string.contains
```

plugins/fish-spec/spec.log.fish
bpinto: :+1:

plugins/fish-spec/test/erase.spec.fish
```diff
@@ -0,0 +1,56 @@
...
+
+  function it_erases_one_element_from_multiple_lists_when_only_one_has_the_element \
+    -d "it_erases_one_element_from_multiple_lists_when_only_one_has_the_element"
+    erase 2 --from nums_until_10 odds_until_10
+    expect $nums_until_10 --to-not-contain 2
+  end
+
+  function it_erases_multiple_elements \
+    -d "it_erases_multiple_elements"
+    erase 1 2 nums_until_10
```
bpinto: Missing `--from`. I do love the `--from` syntax.

spec/spec.fish
```diff
@@ -0,0 +1,80 @@
+import plugins/fish-spec
```
bpinto: This is under `spec` folder.
Are we going to use `spec` or `test`? Both works for me, but we need to use only one for the project.

bpinto: I'd rather not have this file called `spec.fish`, but `oh_my_fish.spec.fish`

plugins/fish-spec/test/erase.spec.fish
bucaran: Got it.

spec/spec.fish
bucaran: I changed it because `oh_my_fish.spec.fish` was too long, and it was annoying it would take almost my entire terminal width. I don't mind to revert it. OK.

bucaran: > Are we going to use spec or test? Both works for me, but we need to use only one for the project.

Definitely `spec` sounds better to me.

spec/spec.fish
```diff
@@ -0,0 +1,80 @@
...
+    expect $fish_function_path --to-contain $fish_path/functions/
+  end
+
+  function it_loads_all_selected_plugins \
+    -d "It should load all user selected plugins"
+
+    erase "$fish_path/plugins/bak" \
+               "$fish_path/plugins/z" --from fish_function_path
+
+    set -g fish_plugins bak z gi
```
bpinto: You added this `gi`, but you forgot to erase it from `fish_function_path` above, so it won't pass on my computer.

I'd remove this `gi` expectation as 2 plugins already guarantees that it loads multiple plugins.

plugins/fish-spec/erase.fish
bpinto: Exciting future! I think this utils library will be a-m-a-z-i-n-g.

spec/spec.fish
bpinto: Okay then. Could you rename the fish-spec test folder?

bpinto: How large is your terminal? :O 
image

bucaran: I love large resolutions, but I use 1024*640 to reduce eye strain.

bucaran: > Okay then. Could you rename the fish-spec test folder?

Yes.

spec/spec.fish
bucaran: Right. Forgot about this one too. Thanks.

spec/spec.fish
bucaran: Last chance to ask, how about `omf.spec.fish`?

plugins/fish-spec/spec.run.fish
bpinto: I've never worked on a test framework that would continue. (e.g. Java, Ruby, Javascript test frameworks)

My opinion is, most of the time a test should only have one expectation. If it has more than one, you might want to take a deeper look and see if it's not testing two things and you couldn't break it into two tests.

If it does need two expectations, one is probably dependent of the other.

Let's consider this test:
```
list.erase 1 nums_until_10
expect $nums_until_10 to_not_include 1
```

There is no guarantees that `1` was included in the list.

```
expect $nums_until_10 --to-contain 1
list.erase 1 nums_until_10
expect $nums_until_10 --to-not-contain 1
```

Now we have, and if the first expectation fails, we don't need to continue the test. As even if it works, it means nothing.

spec/spec.fish
bpinto: I am probably going to convince you to increase your terminal resolution. It's 2015! :+1: 

I don't think we should lose readability of the filename as it is very short. I like that the test filename has the same name as the function. Makes it easier to find the tests, and I can add a script to my vim to switch between the code and the test file. :)

plugins/fish-spec/erase.fish
bucaran: :+1: 

So are we using 
```fish
ls.erase 1 --from $list
```?
or 
```fish
list.erase 1 --from $list
```?

Comment (bpinto) [inline plugins/fish-spec/erase.fish]: I think we should go with `list` and find a way to improve the syntax on your editor.

File: plugins/fish-spec/spec.run.fish (hunk 5 above)
Comment (bucaran) [inline plugins/fish-spec/spec.run.fish]: Tests stop return on failure now. 
:+1:

File: spec/spec.fish (hunk 8 above)
Comment (bucaran) [inline spec/spec.fish]: :+1: 

Hehe.

File: plugins/fish-spec/erase.fish (hunk 1 above)
Comment (bucaran) [inline plugins/fish-spec/erase.fish]: :+1:

File: plugins/fish-spec/spec.log.fish (hunk 4 above)
Comment (bucaran) [inline plugins/fish-spec/spec.log.fish]: You have good powers of persuasion, I now find the `.` way better. Nevermind :laughing:

File: plugins/fish-spec/test/erase.spec.fish (hunk 6 above)
Comment (bucaran) [inline plugins/fish-spec/test/erase.spec.fish]: Fixed.

**Comment by bucaran** [comment]
Going to push.

File: plugins/fish-spec/spec.log.fish (hunk 4 above)
Comment (bpinto) [inline plugins/fish-spec/spec.log.fish]: lol, I wish I had this power... Debating about tests with coworkers at the moment. :P

File: plugins/fish-spec/spec.eval.fish [hunk 10]
```diff
@@ -0,0 +1,96 @@
...
+# OPTIONS
+#   <function>
+#     Name / glob of one or more functions to evaluate. By default find all
+#     functions that match the glob and test each. To run a single function,
+#     use the -u --unit option as described below.
+#
+#   [-u --unit]
+#     Evaluate a single function as indicated by <function>.
+#
+#  [-d --depth <depth>]
```
bpinto: I think `indentation` would be better here. I am still unsure if I understand everything that this guy does.

plugins/fish-spec/spec.run.fish
bpinto: Where was this changed? The comment above is still saying it does not stop.

bpinto: `after_each` and `after_all` should still be called.

plugins/fish-spec/spec.log.fish
bucaran: Co-workers and other humans sleeping here at 1:13 AM. :night_with_stars:

plugins/fish-spec/spec.run.fish
bucaran: Right! OK

plugins/fish-spec/spec.eval.fish
bucaran: You usually see `depth` instead of `indentation` in libraries that create menus, etc. Depth is just the multiplier to generate the indentation. 

`indentation = 2 * depth`

2 is the default indentation, and depth is how many of it you want.

plugins/fish-spec/spec.log.fish
bpinto: :P

plugins/fish-spec/spec.eval.fish
bpinto: Okay.

bucaran:
OK

plugins/fish-spec/spec.eval.fish
```diff
@@ -0,0 +1,96 @@
...
+          set indent (printf " %.0s" (seq (math $depth \* $tab)))
+        else
+          set newline ""
+        end
+      case -s --silent
+        set silent $argv[$index]
+      case -u --unit
+        # Evaluate query as a single function instead of a set.
+        set tests $query
+      case \*
```
bpinto: Is this necessary? As nothing is being done?

plugins/fish-spec/spec.log.fish
```diff
@@ -0,0 +1,59 @@
...
+#
+#   Notice that a message in the form [<color> <string>...] implies that
+#   colors occupy odd and strings even indices after the first option.
+#
+# TODO
```
bpinto: Weird to add this as a TODO, I'd add it as an issue, only if we plan on doing it.

plugins/fish-spec/spec.log.fish
```diff
@@ -0,0 +1,59 @@
...
+  switch $argv[1]
+    case -o --ok
+      set -e argv[1]
+      set argv green . $argv
+    case -f --fail
+      set -e argv[1]
+      set argv red ✘ $argv
+    case -m --msg --message
+      # Default message color.
+      set argv[1] white
```
bpinto: I think this does not allow a color to be specified when calling `-m`.

plugins/fish-spec/spec.run.fish
```diff
@@ -0,0 +1,66 @@
...
+    # return 1 if any test do fails.
```
bpinto: I think this message could be rewritten.

> Break execution if the test fails.

bpinto:
Almost there.

plugins/fish-spec/spec.log.fish
bucaran: That option was not available from the start. Of course, we could add it later.

bpinto:
```
#   -m --msg --message
#     Output the message according to the <color> <string>... list.
```

plugins/fish-spec/spec.eval.fish
bucaran: Fixed.

plugins/fish-spec/spec.log.fish
bucaran: Yes, tricky. Only the first message is white to allow:

`spec.log "Error" yellow this red that green here!`

plugins/fish-spec/spec.run.fish
bucaran: Done.

plugins/fish-spec/spec.log.fish
bucaran: Done.

plugins/fish-spec/spec.log.fish
bucaran: Let me add this.

bpinto: Wants to do this after we merge this pull request?

I think we need a `log` library as part of the `utils` plugin.

bucaran:
Are we good?

bpinto:
Awesome awesome!

Issue #299, opened 2015-01-10
import function shouldn't add functions defined in the test folder to path


bpinto:

bucaran:
What do you mean? `import` has no knowledge of any tests.

bpinto:
This is in my `$fish_function_path`: `/Users/bpinto/.oh-my-fish/plugins/dpaste/tests/`.
There is no reason to add test files to this environment variable. Import could be smart about it.

bucaran:
Ah, I see what you mean. Yes, I knew about this. How do you want to fix this?

1. Pass a list of directories that should not be added regardless of their contents to `import`.
2. Or just let `import` be smart and not add `test/` or  `tests/` directories.

bpinto:
I would go with the second option for now. I think we don't need anything fancy atm.

bucaran:
I'll PR this in a `fix-import` branch after I am done with tests and travis stuff.

bpinto:
:+1:

bpinto:
Thanks @bucaran !

bpinto:
Thanks @bucaran !

bucaran:
The culprit is actually `_prepend_tree`. Fixed.

bucaran:
Let's continue over #308

PR #308, opened 2015-01-13
Do not add test/spec files to function path in _prepend_tree.

bucaran:
Do not add test/spec files to function path in _prepend_tree.

### DIFF
--- a/functions/import.fish
+++ b/functions/import.fish
@@ -12,7 +12,8 @@
 # NOTES
 #      $fish_path and $fish_custom point to oh-my-fish home and the user
 #      dotfiles folder respectively. Both globals are usually configured
-#      in ~/.config/fish/config.fish
+#      in ~/.config/fish/config.fish. Also, import is clever enough to
+#      skip directories with *.spec.fish files.
 #
 # EXAMPLES
 #      import plugins/<plugin>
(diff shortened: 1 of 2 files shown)


bpinto:
I think the best place to change is on the `import` function. Can't we pass a glob to _prepend_tree that will not add `.spec.fish`?

bucaran:
Yes.

bpinto:
I think we should clean the other plugins, otherwise their test are not going to be executed on travis, which would be bad.

bucaran:
Exactly. All plugins should conform to the new spec guidelines (we have to write something up).

bpinto:
I fix the plugins and you write the guideline? I know you love to write. :+1:

bucaran:
:+1: 

Does that belong to the repo? or to the wiki?

bpinto:
I think one thing we need to start doing is improving our documentation.

We are adding documentation to the files now, but we have no `README.md` on the plugins. It would be helpful, specially for `fish-spec`.

I usually don't like wiki, as I think people don't read, at least we need a small description on the repo readme with a link to the wiki page.

functions/import.fish
```diff
@@ -25,14 +26,17 @@
...
 function import -d "Load libraries, plugins, themes, etc."
+  # Do not add spec files to function path.
+  set -l skip \*.fish -a ! \*.spec.fish
+
   for library in $argv
     # Prepend plugins, themes and completions, traversing library
     # trees and prepending directories with fish code.
-    _prepend_tree $fish_path/$library
-    _prepend_tree $fish_custom/$library
+    _prepend_tree $fish_path/$library     $skip
```
bpinto: `$skip_tests`?

bucaran: We may want to skip something else in the future maybe. You seem to like long variable names, and I kind of always prefer one word names in order to avoid using `_`.

bpinto: Future is future. :)

Long names, I wouldn't say so, I prefer descriptive names. `$skip` means nothing. :)

bucaran: Done.

bpinto:
Want to rebase this as I have merged that other PR?

bpinto:
Fixed #299

bpinto:
Closes #299

Issue #307, opened 2015-01-13
Subsequent tests should run if one test fails


bpinto:
```
.oh-my-fish:run_plugins_tests_on_travis* λ ./script/run-tests.fish
...................⏎                                                                            

.oh-my-fish:run_plugins_tests_on_travis* λ ./script/run-tests.fish
.............✘
Expected/Users/bpinto/.oh-my-fish/custom
to not contain/Users/bpinto/.oh-my-fish/custom⏎
```

Note that after one test failed, other tests were not executed.

bucaran:
You said you wanted to avoid this, so I deliberately added this functionality.

bpinto:
I think I wasn't clear then, I think tests should be isolated, they could even run in parallel, so one test should not affect the other.

But during one test execution, if there is an error on an expect, it should halt the test execution at that point.

bucaran:
> But during one test execution, if there is an error on an expect, it should halt the test execution at that point. But the test suite should not stop.

You mean stop that one test function, but not the other tests in the suite?

That's exactly how it was before... :laughing:

bpinto:
Sorry about that. But I thought I had tested and it didn't work... =/

bucaran:
Mmm, one solution would be using a global variable with an event handler tracking failed `expect` calls.

bpinto:
I haven't thought about it yet, need to do some testing before. I'll work on it soon, but after the other issues.

bucaran:
So, first thing, let me revert it so that test don't stop when one fails, but rather continue until the end of the suite and we can figure out a way to stop single tests if one of the `expect` calls fail.

What do you think?

bpinto:
Sounds good!

bucaran:
Well, I am working already on the whole thing as I realized there is an easier solution without event handlers. While at it, I discover there was a bug in the initial setup script.

bucaran:
Where are we?

PR #309, opened 2015-01-13
Make sure to exit with 0 of 1 when reloading the framework

bucaran:
Make sure to exit with 0 of 1 when reloading the framework.

This paves the way for [this](https://github.com/bpinto/oh-my-fish/issues/307#issuecomment-69794795) too.

### DIFF
--- a/spec/oh-my-fish.spec.fish
+++ b/spec/oh-my-fish.spec.fish
@@ -69,6 +69,13 @@ function describe_oh_my_fish -d "Oh-My-Fish test spec..."
     load_oh_my_fish
     expect $fish_function_path --to-contain $fish_path/themes/l
   end
+
+  function it_reloads_with_status_of_0 \
+    -d "It should reload the framework with \$status of 0"
+
+    load_oh_my_fish
+    expect $status --to-equal 0
+  end
 end
 
 function load_oh_my_fish
--- a/oh-my-fish.fish
+++ b/oh-my-fish.fish
@@ -29,3 +29,6 @@ end
 
 # Prepend extracted user functions so they have the highest priority.
 set fish_function_path $user_function_path $fish_function_path
+
+# Make sure to exit with $status of 1 when reloading the framework.
+or true


bpinto:
I don't understand this, `set fish_function_path $user_function_path $fish_function_path` should return always true, right?

P.S.: Can we add a test for this?

bucaran:
This is necessary to allow halting individual test execution. What happens is simply that the 

```fish
if not contains $fish_path/functions/ $fish_function_path
  set fish_function_path $fish_path/functions/ $fish_function_path
end
```
When you reload the session, $fish_path is already added to $fish_function_path,  so the condition modifies the `$status` variable causing the reload to exit with 1.

Try using a theme that shows a visual cue like changing the color of the prompt after a command runs like _bobthefish_ and try reloading `. $fish_path/oh-my-fish.fish` and you will see.

bpinto:
I have my head wrapped on something else for now, but adding a test case would be nice.

bucaran:
What kind of test?

bpinto:
That running `. $fish_path/oh-my-fish.fish` returns 0.

bucaran:
Ahh, sure, that should be simple to add.

bucaran:
Done.

bpinto:
Does the test fail if you remove your code?

bucaran:
Testing....
__Yes__.

yes

Merge order should be:
+ reload-and-exit-ok
+ stop-test-continue-suite

bpinto:
Awesome! If we add a test for every bug we found, this will be amazing!

My goal is to start adding test for the `fish-spec` plugin, it's a little bit harder, but required.

spec/oh-my-fish.spec.fish
```diff
@@ -69,6 +69,13 @@ function describe_oh_my_fish -d "Oh-My-Fish test spec..."
     load_oh_my_fish
     expect $fish_function_path --to-contain $fish_path/themes/l
   end
+
+  function it_reloads_with_status_of_0 \
+    -d "It should reload the framework with $status of 0"
```
bpinto: Shouldn't this $status be escaped?

bucaran:
> My goal is to start adding test for the fish-spec plugin, it's a little bit harder, but required.

Absolutely.

spec/oh-my-fish.spec.fish
bucaran: Yes. Done.

bpinto: Where? :P

bucaran: Sorry, I forgot to `-f` force the push since I squashed the \$ bug.

bpinto:
@bucaran You are still using "should", I'll convince you. :)

bucaran:
You mean for the long descriptions? Where where? :laughing:

bpinto:
:P

PR #310, opened 2015-01-13
Stop test continue suite

bucaran:
There you go. Individual tests stop running if an expect fails or _anything fails inside the test function_, but the whole suite continues until the end.

### DIFF
--- a/plugins/fish-spec/expect.fish
+++ b/plugins/fish-spec/expect.fish
@@ -21,6 +21,12 @@
 #   spec.run
 #/
 function expect
+  # Abort if last call to `expect` finished with $status 1. This allows to
+  # stop individual tests from running if at least one expect call fails.
+  if [ $status -eq 1 ]
+    return 1
+  end
+
   set -l expected  $argv[1..-3]
   set -l condition $argv[-2]
   set -l actual    $argv[-1]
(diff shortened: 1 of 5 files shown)


plugins/fish-spec/expect.fish
```diff
@@ -21,6 +21,12 @@
 #   spec.run
 #/
 function expect
+  # Abort if last call to `expect` finished with $status 1. This allows to
+  # stop individual tests from running if at least one expect call fails.
+  if [ $status -eq 1 ]
```
bpinto: It seems like the `expect` need to be in chain for this to work:

Works:
```
expect 'a' --to-equal 'a'
expect 'a' --to-equal 'a'
```

Does not work:
```
expect 'a' --to-equal 'a'
echo 'abc'
expect 'a' --to-equal 'a'
```

bucaran: Ah certainly. This hack will break the implementation because `echo` sets the `$status` to 0.

bpinto:
This is an improvement, but bash or javascript, a broken expectation needs to make the test as a failure, independent of what was last called.

bucaran:
Did you merge this?

PR #311, opened 2015-01-13
Modifying bak plugin tests to use fish spec

bpinto:
Also improved `--to-contain` to check multiple individual arguments.


bpinto:
One test is breaking, but I can't reproduce locally.

Need to improve error handling of `fish-spec`.

bucaran:
Which one?

bpinto:
It was failing before, I think I fixed in a previous commit to `expect`. Want to review this PR?

bucaran:
Let's fix `expect`. I am trying to reproduce the test that you mentioned was failing:

```fish
expect 'a' --to-equal 'b'
```

bpinto:
What do you mean, fix `expect`?

bucaran:
Nevermind!

PR #313, opened 2015-01-14
WIP: Fish spec improvements

bpinto:


bpinto:
@bucaran you might want to see this.

bucaran:
Good stuff. Let me see.

plugins/fish-spec/spec/suite.spec.fish
```diff
@@ -0,0 +1,101 @@
+import plugins/fish-spec
```
bucaran: Tests testing themselves?

plugins/fish-spec/spec.run.fish
```diff
@@ -41,11 +41,15 @@ function spec.run
...
   end
-  # Run 1 or more topmost describe blocks causing fish to load inmediately
+
+  # Run 1 or more topmost describe blocks causing fish to load immediately
   # scoped functions, i.e, those inside any invoked describe_ blocks.
   spec.eval describe_ --depth 0 $output
   spec.eval before_all $output
-  for test in (spec.functions "it_")
+
+  set -l tests_names (spec.functions "it_")
```
bucaran: :+1: Long names.

bpinto: lol, really? :P

plugins/fish-spec/spec.run.fish
```diff
@@ -56,7 +60,12 @@ function spec.run
     # Make sure to run after_each even if a test fails.
     spec.eval after_each $output
   end
+
   spec.eval after_all $output
 
+  if [ -z "$tests_names" ]
+    echo 'No tests found.'
```
bucaran: Another option is to return > 1 here, and leave `spec.run` caller to deal with this.

2: _Tests not found error_

plugins/fish-spec/spec/suite.spec.fish
bpinto: This was hard to implement, but I'm glad with the end result, I think the tests are as simple as they could be.

plugins/fish-spec/spec.run.fish
bucaran: \*`test_names`

bpinto: I never know, will fix it.

plugins/fish-spec/spec/suite.spec.fish
bucaran: Awesome.

plugins/fish-spec/expect.fish
```diff
@@ -43,6 +43,11 @@ function expect
     end
   end
 
+  if test -z "$expected"
+    spec.log --fail red \n"Expect called without an expectation argument."
+    return 1
```
bucaran: Excellent.

plugins/fish-spec/spec.run.fish
bpinto: I don't know what you mean by let the `spec.run` caller deal with this.

```
# Begin test suite ❯❯❯
spec.run $argv
```

This is the caller.

bucaran: Or just:

```fish
  if [ -z "$tests_names" ]
    spec.log --fail red \n"spec.run called, but no tests were found."
    return 1
  end
```

plugins/fish-spec/expect.fish
bpinto: I don't think this check is working anymore.

bucaran: Same here. In the meantime I am trying to fix `spec.eval` so that it will print newlines after test suites, but not to the first describe of a series of suites. I had originally used `--depth 0` for that, but it feels more like a hack.

bpinto: I have deleted it, will do this in the future.

bucaran: It seems anything I print inside spec.run fails with

```
Expected Initializing fish-spec...
to equal.✘
Expected Initializing fish-spec...
to equal..✘
Expected Initializing fish-spec...No tests found.
```

plugins/fish-spec/spec.run.fish
bpinto: It is not a failure, just a message, it could be running multiple suites.

plugins/fish-spec/expect.fish
bucaran: Fixed if moved inside `if contains -- $argv[1] -v --verbose` block, where it should since any echoing should go into --verbose mode.

bpinto: I am lost, are you talking about the code I removed?

bucaran: Yeah. I made a small change. Please see.

bpinto: It should be added in a future PR with a test for it. No need to hold this PR for another feature.

bucaran:
```fish
#!/usr/bin/env fish

for test in (find * -type f -print | grep 'spec.fish')
  fish $test $argv
end
```

bpinto: :+1:

plugins/fish-spec/spec.run.fish
bucaran: Well `--message`?

bpinto:
We need to work on #314 and #315 before we continue this.

bucaran:
Here is an idea. Let me try giving it a shot using `msg` instead of `spec.log`? What you say?

bpinto:
Done on #326

Issue #315, opened 2015-01-14
Test output needs improvement


bpinto:
```
'Testing a suite run'..Array index out of bounds
~/.oh-my-fish/plugins/fish-spec/expect.fish (line 10):       set -g expected  $argv[1..(math "$i - 1")]
                                                                                        ^
in function “expect”
	called on standard input
	with parameter list “--to-equal No tests found.”
in function “it_has_an_output_if_suite_is_blank”
	called on line 1 of file -
from sourcing file -
	called on line 52 of file /usr/share/fish/functions/eval.fish
in function “eval”
	called on line 94 of file ~/.oh-my-fish/plugins/fish-spec/spec.eval.fish
in function “spec.eval”
	called on line 52 of file ~/.oh-my-fish/plugins/fish-spec/spec.run.fish
	with parameter list “it_has_an_output_if_suite_is_blank --unit --depth 1 ”
in function “spec.run”
	called on line 121 of file ~/build/bpinto/oh-my-fish/plugins/fish-spec/spec/suite.spec.fish
	with parameter list “-v”
✘
Expected..
to equalNo tests found...'Testing Oh-My-Fish `list.erase` function...'
  'It should erase multiple elements.'..
  'It should erase multiple elements from multiple lists.'...
  'It should erase multiple elements with --from option.'..
  'It should erase an element from a list.'.
  'It should erase an element from multiple lists.'..
  'It should erase an element from multiple lists when only one has the element.'.
  'It should erase an element with --from option.'.'Testing bak plugin'..........✘
Expecteda/ a.20150114_021644.bak/
to equala.20150114_021644.bak/ a/....'Oh-My-Fish test spec...'
```

Ref: https://travis-ci.org/bpinto/oh-my-fish/builds/46936755

bucaran:
Here is an idea. Let me try giving it a shot using msg instead of spec.log? What you say?

bpinto:
I think it could be a good solution.

bucaran:
Will work on it right away!

bucaran:
Any preferences regarding UI?

I like this http://unitjs.com/assets/img/screen_console_unitjs_and_mocha_demo2.png

bpinto:
I like that UI a lot, it's clean.

I don't know what happens when a test fails, but I think the error message should be displayed after all tests have run.

bpinto:
[...]

Issue #314, opened 2015-01-14
run-tests.fish should exit with status != 0 if a test failed.


bpinto:
```
$ script/run-tests.fish
......................✘
Expecteda/ a.20150114_020353.bak/
to equala.20150114_020354.bak/ a/............
The command "script/run-tests.fish" exited with 0.
```

From: https://travis-ci.org/bpinto/oh-my-fish/builds/46935945

bucaran:
Ok, I know why the tests were failing. I actually allowed this to happen as I wasn't sure how this should be handled and forgot to bring it up.

Currently, only `it_` kind of tests are allowed to fail the suite, `before_all`, etc., do not even if they fail. 

I figured why would anyone add an expect condition inside a `before_`, `after_` type of tests...

bpinto:
Fixed by #326

PR #317, opened 2015-01-14
Bak descriptions et al

bucaran:
Add proper descriptions, remove _should_, etc.

### DIFF
--- a/plugins/bak/spec/bak.spec.fish
+++ b/plugins/bak/spec/bak.spec.fish
@@ -1,7 +1,7 @@
 import plugins/bak
 import plugins/fish-spec
 
-function describe_bak -d 'Testing bak plugin'
+function describe_bak -d 'bak plugin'
   function before_all
     set -g test_dir /tmp/bak_test
     mkdir -p $test_dir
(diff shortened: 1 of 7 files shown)


bpinto:
Same as #313 ; we can only push it after #314 and #315.

bucaran:
In that order?

bpinto:
#315 would help solving #314. :)

bucaran:
Alright. Will work on this first.

bucaran:
Some of the changes that I will be pushing:

+ Remove spec.log and use `msg`.
+ Everything will be silent by default, but you can turn verbose optionally.
+ spec.eval will no longer evaluate multiple functions as there is no need. The main loop occurs in spec.run. 
+ An improved UI. Similar to mocha. With a failed tests report after the run.
+ Will add some more options to spec.eval such as `--title` to indicate the current function should be displayed as a header. These options should be semantic, as spec.run should not deal with any design issues such as colors, etc.
+ expect will no longer print anything. Instead it will notify the results back to spec.eval where output will be displayed accordingly and based on the --verbose flag.

Any comments?

plugins/bak/spec/bak.spec.fish
```diff
@@ -18,24 +18,25 @@ function describe_bak -d 'Testing bak plugin'
     popd
   end
 
-  function it_checks_bak_filename_pattern_is_followed
+  function it_checks_bak_filename_pattern_is_followed \
+    -d "checks `bak` filename pattern is followed"
```
bucaran: followed by what?

bpinto: If we stick to the pattern. I guess. :P

bucaran: :+1:

bpinto:
> Everything will be silent by default, but you can turn verbose optionally.

Don't know what that means, but if by default it only returns status 0 or 1, I think this is not expected. I'd have a `--quiet` option for this instead.

I think by default it should output

```
......X.....X....

fish-spec/list.erase.fish erases an element failed:
Expected: X
Actual: y
```

bucaran:
Edited.

bpinto:
Sounds reasonable. :+1:

bucaran:
:+1: OK

bucaran:
I will call the PR `fish-spec-upgrade`. See you in a few hours.

bucaran:
> The above implies that there must be only one describe, before_all after_all, etc., per test suite. This will probably be good enough for now, but we should improve this in the future.

bpinto:
:+1:

bucaran:
We should be able to have multiple describes if they are not nested. Would rather have multiple files for this, though.

bucaran:
```fish
  function it_has_an_output_if_suite_is_blank
    set -l suite "
        import plugins/fish-spec

        function describe_blank_suite
        end

        spec.run
      "

    expect (run_nested_suite $suite) --to-equal 'No tests found.'
  end
```
Why? What is the advantage of this?

bpinto:
Multiple describes if they are not nested is something we should support as we should support nested describes. I know nested describes is only coming in the future, but no reason to remove multiple describe blocks, if that is what you were thinking.

Specs without describe blocks should also output "No tests found."

PR #319, opened 2015-01-15
Fish spec upgrade.

bucaran:
#315  #317 

+ Add fish eval fix where eval would modify the $status if the evaluated
  expression does not modify $status itself.
+ New output system with colors and other niceties.
+ Improvements to expect function removing unneded code.
+ Add correct support for multiple non-nested describe blocks within the
  same suite.
+ Remove spec.log and favor msg and standard echo.
+ Add support to print errors to stderr and test output to stdout allowing
  test runners (e.g. spec.run) to customize the order in which the output
  is presented. the current version displays error dumps under each test,
  but it would be possible to implement now without modifying the core.
+ spec.eval no longer evaluates multiple functions as there is no need
  since the main loop is already handled in spec.run.
+ spec.eval now displays meta information about the test such as total of passed
  and failed tests, and displaying error dump from stdout.
+ etc


bucaran:
`oh-my-fish`, `list.erase`, tests + new additions pass, older tests mostly fail due to the new commits.

bucaran:
The UI was greatly improved and the code is still manageable. Support was added to stderr and stdout output so modifying the order in which the test output is presented is just a matter of tweaking things around. I originally tried printing the dump after all messages, but didn't like how it looked and prefer the current design.

untitled

bpinto:
This looks neat! I won't be able to review this in less than 4 hours though =/

bucaran:
I will keep adding, squashing and pushing until for another while. Looking forward to your review.

plugins/fish-spec/expect.fish
```diff
@@ -23,58 +28,63 @@
...
     case --to-be-true
       eval "$expected"
-      test $status -eq 0
-    case --to-contain
-      set result 0
+    case --to-contain --to-not-contain
       for item in $actual
         contains -- "$item" $expected
-          or set result $status
+          and break
```
bpinto: This is weird, is this rebased with master? Yesterday I added tests for the `expect` function.

plugins/fish-spec/expect.fish
```diff
@@ -23,58 +28,63 @@
...
-          or set result $status
+      switch $status
+        case 0
+          test $condition = --to-contain
+        case \*
+          test $condition != --to-contain
       end
-      test $result -ne 0
-    case --to-eq\*
+    case --to-eq\* --eq -eq =
```
bpinto: I don't like these alias, it's unreadable.

`expect 'a' = 'b'`. Does not read the same as `expect 'a' --to-equal 'b'`.

plugins/fish-spec/expect.fish
```diff
@@ -23,58 +28,63 @@
...
+    case --to-no-eq\* --ne -ne !=
+      set condition --to-not-equal
+      test "$expected" != "$actual"
   end
+
   set result $status
   if [ $result -eq 0 ]
-    spec.log --ok
+    # Print a non-empty string to indicate our spec parser a test succeded.
+    printf "$result"
```
bpinto: I thought `expect` was no longer printing anything.

plugins/fish-spec/spec.eval.fish
```diff
@@ -1,98 +1,108 @@
...
-  # Keep a copy in case -u --unit option is used.
-  set -l query $argv[1]
-  # Collect all functions that start with $query's key.
-  set -l tests (spec.functions $query)
-  set -l tab      2
-  set -l indent  ""
-  set -l newline \n
-  set -l silent  ""
+# Redefine these colors to customize the report.
+set -g spec_test_pass_color @greenyellow
```
bpinto: These variables should be defined inside the function so they are self-contained.

plugins/fish-spec/spec.eval.fish
```diff
@@ -1,98 +1,108 @@
...
-      case -s --silent
-        set silent $argv[$index]
-      case -u --unit
-        # Evaluate query as a single function instead of a set.
-        set tests $query
+function spec.eval
+  set -l result $status
+  set -l info   ""
+  set -l test   $argv[1]
+  if functions -q $test
```
bpinto: Empty lines would improve readability of this file.

plugins/fish-spec/spec.eval.fish
```diff
@@ -1,98 +1,108 @@
...
+    set -e argv[1]
+    # Parse options
+    set -l tab    1
+    set -l verb  ""
+    set -l title ""
+    set -l keep  ""
+    for index in (seq (count $argv))
+      switch "$argv[$index]"
+        case -v --verbose
+          set verb -v
```
bpinto: `verb`?

plugins/fish-spec/spec.functions.fish
```diff
@@ -1,16 +1,21 @@
...
+#
+# OPTIONS
+#   See grep(1)
 #
 # EXAMPLES
 #   spec.functions "describe_"
-#   spec.functions "before_"
+#   spec.functions -q "before_"
 #/
 function spec.functions -a key
```
bpinto: Arent you using a named parameter?

plugins/fish-spec/spec.run.fish
```diff
@@ -31,32 +32,60 @@
 #     end
 #     spec.run --verbose
 #
+# NOTES
+#   import plugins/spec
```
bpinto: Where it used to be reads better.

plugins/fish-spec/spec.run.fish
```diff
@@ -31,32 +32,60 @@
...
-  end
-  # Run 1 or more topmost describe blocks causing fish to load inmediately
-  # scoped functions, i.e, those inside any invoked describe_ blocks.
-  spec.eval describe_ --depth 0 $output
-  spec.eval before_all $output
-  for test in (spec.functions "it_")
-    spec.eval before_each $output
+  set -l result  0
+  set -l tests  ""
+  set -l suites (spec.functions describe_)
```
bpinto: `describes`

plugins/fish-spec/expect.fish
bucaran: It is rebased. I made some changes expect. Mostly cutting unneeded lines.

plugins/fish-spec/expect.fish
bucaran: Me neither! and I don't plan to use them myself either, but they are there for consistency with fish. It's how fish itself rolls.

plugins/fish-spec/spec.run.fish
```diff
@@ -31,32 +32,60 @@
...
-      set result 1
+    for test in $tests
+      spec.eval before_each --keep $argv
+      # Capture error dump coming from stdout in case test fails.
+      if not set -l stdout (spec.eval $test --tab 1 $argv)
+        # Flunk if any single test fails, but do not stop the suite.
+        set result 1
+        # Keep fail count.
+        set failed (math 1+$failed)
+        # Build dump output to present after count report.
```
bpinto: What do you mean?

plugins/fish-spec/expect.fish
bucaran: Yes, but these are just being set to stdout and later caught by `spec.eval`, these do not end up in the terminal, but are processed by `spec.eval`. Only expect knows what went right or wrong and the diagnosis.

plugins/fish-spec/spec.eval.fish
bucaran: What function? There's no function anymore.

plugins/fish-spec/spec.eval.fish
bucaran: :+1:

plugins/fish-spec/spec.eval.fish
bucaran: `verbose`.

plugins/fish-spec/expect.fish
bpinto: Where have you seen this?

plugins/fish-spec/spec.functions.fish
bucaran: It could be useful to query for functions similar to `functions -q`. There's nothing new here, I am just passing $argv along to grep.

plugins/fish-spec/expect.fish
bpinto: Makes sense.

plugins/fish-spec/expect.fish
```diff
@@ -23,58 +28,63 @@
...
   else
-    spec.log --fail red \n"Expected" yellow "$expected" \
-                    red \n(echo $condition | tr "-" " " \
-                                           | cut -c 3-) yellow "$actual"
+    # Print the expected, actual and condition in title case to the stdout.
+    set condition (echo $condition | tr "-" " " | cut -c 3- \
```
bpinto: This is crazy, we need tests for this.

plugins/fish-spec/spec.eval.fish
bpinto: There is a function below `spec.eval`

plugins/fish-spec/spec.eval.fish
bpinto: Rename? :+1:

plugins/fish-spec/expect.fish
bucaran: I mean, it's fish same syntax. Like, `test a -eq b`.

plugins/fish-spec/spec.functions.fish
bpinto: But the line below resets `key`.

plugins/fish-spec/expect.fish
bucaran: `printf "$result"` Result could be anything really as long as it is not an empty string. This is how we tell `spec.eval` that the test was successful.

plugins/fish-spec/spec.eval.fish
bucaran: If you mean `spec.get.info` it's gone. I figured, we are using this only once. Actually, you figured that first :laughing:

plugins/fish-spec/spec.eval.fish
bucaran: :feelsgood: :+1:

plugins/fish-spec/spec.run.fish
bucaran: I need to fix this. I forgot to add a note about `spec` having to be imported via `import` in order to use it.

plugins/fish-spec/spec.run.fish
bucaran: :+1:

plugins/fish-spec/spec.run.fish
bucaran: I forgot to remove that comment. Originally, I had the error dump / stack trace / diagnosis portion of the output _after_ all output, but couldn't agree on a design so it's how it is now. 

But the possibility is there to be implemented later.

plugins/fish-spec/expect.fish
bucaran: That's `title case`. Sure.

That's what this is all about!

plugins/fish-spec/spec.functions.fish
bucaran: Typo. Thanks.

plugins/fish-spec/expect.fish
bpinto: Fish or `test` syntax?

I don't know... I feel like this is a bad call. We can do better than supporting `eq` and `ne`.

plugins/fish-spec/spec.eval.fish
bpinto: All I am saying is to define these variables inside the function that uses it. :P

plugins/fish-spec/expect.fish
bucaran: What do you suggest? I could remove them, as no one should be using them anyway, but not having them feels inconsistent in shellspeak.

plugins/fish-spec/spec.eval.fish
bucaran: Ah, you the global variables? These are used across the application. Maybe adding a file with the declarations would be better?

plugins/fish-spec/expect.fish
bpinto: I think we are already using `fish`, which is different and supposed to be more advanced than old shells.
I feel like good code have a very good readability, and we can help them write better tests by not having these old-school things. :)

plugins/fish-spec/spec.eval.fish
bpinto: I couldn't find its usage anywhere but inside the function `spec.eval`.

plugins/fish-spec/spec.run.fish
bpinto: I think it's god to have them inline when we are running in verbose mode.
I will have a play with it later though.

spec/oh-my-fish.spec.fish
```diff
@@ -71,8 +69,8 @@ function describe_oh_my_fish -d "Oh-My-Fish test spec..."
   end
 
   function it_reloads_with_status_of_0 \
-    -d "It should reload the framework with \$status of 0"
-
+    -d "reloads the framework with \$status of 0"
+    echo original
```
bpinto: Debugging?

plugins/fish-spec/expect.fish
bucaran: Off they go then. Users of `fish-spec` will be happy.

plugins/fish-spec/spec.eval.fish
bucaran: They are used in expect and spec.run. too.

plugins/fish-spec/spec.run.fish
bucaran: Alright.

spec/oh-my-fish.spec.fish
bucaran: Lol. Fixed. Thanks

plugins/fish-spec/spec.run.fish
bucaran: Fixed.

bucaran:
BTW, some new tests do not have any descriptions. It doesn't have to be perfect, but tests should have descriptions, it's a _feature_, let's use it.

plugins/fish-spec/expect.fish
bpinto: I can't understand this break, it seems to me that not all items are being checked

spec/oh-my-fish.spec.fish
bpinto: Ping me when you push your latest, code looks good!

plugins/fish-spec/expect.fish
bucaran: I see now that what you want is to check if it contains ALL the elements. Fixed.

bpinto: I'm confident I had written a test that checks this specifically, did you run the tests?

One thing at a time, there are many many extra expects we could add, but we don't need to increase the scope of this pull request, the bigger it is, the harder it is to review and the longer it takes to be merged.

bpinto:
I am not happy with how we have to write the function name and the description with almost if not equal text. We should be able to parse the function name and obtain an output text to use if no description message was specified.

bpinto:
For this PR scope we can just print the function name and in a future PR we can do this function name parser.

plugins/fish-spec/expect.fish
bucaran: :+1:

bucaran:
I agree and it's not hard, the thing is then though, spec.eval would need skip some output from `before_*` and `after_*` which I thought was inconsistent, but spe

bucaran:
> For this PR scope we can just print the function name and in a future PR we can do this function name parser.

No need for that. It's _done_. Now it uses the description _only_ if there's one. This way, if the user really wants to use a custom description, they still can.

bpinto:
:+1: nice @bucaran

plugins/fish-spec/expect.fish
```diff
@@ -1,15 +1,20 @@
...
-#            --to-contain       <actual> values exist in <expected> list
-#            --to-no-contain    <actual> values does not exist in <expected> list
-#            --to-equal         <actual> value equals <expected> value
-#          <actual>
+#   expect <expected> <condition> <actual>
+
+# OPTIONS
+#   <expected>...
+#   <condition>
+#     --to-equal       -eq  <actual> value equals <expected> value
```
bpinto: `-eq` and `-ne` support is gone! :

bucaran:
I just found out that, the spec spec is failing because of the color control codes added to the strings, so when expect compares "A" with "A", what really gets matched is something like "[;24mA" and "A", etc. I am in the process of removing these so that all tests can pass.

bpinto:
Maybe on the tests we can use `grep -E \w*` to only fetch the text? Don't know...

plugins/fish-spec/spec/expect.to_be_false.spec.fish
```diff
@@ -1,16 +1,16 @@
 import plugins/fish-spec
 
-function describe_to_be_false -d 'expect --to-be-false'
+function describe_fish-spec:expect_to_be_false
```
bpinto: I find this syntax very weird. I'd rather use `-d` instead.

plugins/fish-spec/spec/expect.to_be_true.spec.fish
```diff
@@ -1,6 +1,6 @@
 import plugins/fish-spec
 
-function describe_to_be_true -d 'expect --to-be-true'
+function describe_fish-spec:expect_to_be_true
```
bpinto: We are always inside `fish-spec`, no need to say it on every suite.

plugins/fish-spec/spec/list.erase.spec.fish
```diff
@@ -1,56 +1,68 @@
...
 
-function describe_list_erase -d "Testing Oh-My-Fish `list.erase` function..."
+function describe_fish-spec:list.erase
   function before_each
     set -g nums_until_10 1 2 3 4 5 6 7 8 9 10
     set -g odds_until_10 1 3 5 7 9
   end
 
-  function it_erases_one_element -d "It should erase an element from a list."
+  function after_each
```
bpinto: Blank after each?

plugins/fish-spec/spec/list.erase.spec.fish
```diff
@@ -1,56 +1,68 @@
...
+  end
+
+  function it_erases_one_element
     list.erase 1 nums_until_10
     expect $nums_until_10 --to-not-contain 1
   end
-
-  function it_erases_one_element_without_from_option \
-    -d "It should erase an element with --from option."
+  #
```
bpinto: Left overs?

plugins/fish-spec/spec/list.erase.spec.fish
```diff
@@ -1,56 +1,68 @@
...
+    list.erase 10 --from nums_until_10
+    expect $status --to-equal 0
+  end
+
+  function it_returns_1_if_no_items_are_erased
+    list.erase 100 200 300 --from nums_until_10
+    expect $status --to-equal 1
+  end
+
+  function it_returns_1_if_no_items_are_erased_from_any_lists
```
bpinto: What happens if it erases from some but not all lists? Can we have a test for this scenario?

spec/oh-my-fish.spec.fish
```diff
@@ -1,6 +1,6 @@
-import plugins/fish-spec
+import plugins/fish-specs
```
bpinto: No plural, `plugins/fish-spec`

plugins/fish-spec/expect.fish
```diff
@@ -23,27 +28,27 @@
...
+  set -l expected  ""
+  set -l condition ""
 
-      # No comparison required e.g. --to-be-true
-      if not [ (count $argv) = $i ]
-        set -g actual $argv[(math "$i + 1")..-1]
-      end
-
-      break
+  for index in (seq (count $argv))
```
bpinto: :+1: index

plugins/fish-spec/spec.eval.fish
```diff
@@ -1,98 +1,112 @@
...
+      set info (echo $test | tr "_" " ")
+    end
+    
+    set -e argv[1]
+    # Parse options
+    set -l tab       1
+    set -l verbose  ""
+    set -l title    ""
+    set -l keep     ""
+    for index in (seq (count $argv))
```
bpinto: $argv count can be 0.

```
if [ -n "$argv" ]
```

plugins/fish-spec/spec/suite.spec.fish
```diff
@@ -0,0 +1,121 @@
...
+    set -l suite "
+        import plugins/fish-spec
+
+        function describe_blank_suite
+        end
+
+        spec.run
+      "
+
+    expect (run_nested_suite $suite) --to-equal 'No tests found.'
```
bpinto: Use `--to-contain` instead.

plugins/fish-spec/spec/suite.spec.fish
```diff
@@ -0,0 +1,121 @@
...
+  end
+
+  # function it_runs_all_describe_blocks
+  #   set -l suite "
+  #       import plugins/fish-spec
+  #
+  #       function describe_blank_suite
+  #         echo 'first describe'
```
bpinto: I think we need to use signal handling to handle this.

bucaran:
I am about to push the latest changes. Now there's a `spec.view` method that handles all output, relieving `spec.run` and `spec.eval` from this.

plugins/fish-spec/spec/expect.to_be_false.spec.fish
bucaran: You mean the `:`? :+1:

plugins/fish-spec/spec/expect.to_be_true.spec.fish
bucaran: There is no nesting so it helps visualize it, but they do happen orderly so you may be right.

plugins/fish-spec/spec/list.erase.spec.fish
bucaran: Leftovers.

spec/oh-my-fish.spec.fish
bucaran: So it was silently failing to load the plugin, but since it was already loaded in the session it didn't complain. Ha.

bucaran:
BTW using `grep -E` will not work on Linux, so avoid it or check with `uname` and provide different solutions. I'd rather avoid for the time being.

plugins/fish-spec/spec/expect.to_be_false.spec.fish
bpinto: Yep

plugins/fish-spec/spec/expect.to_be_true.spec.fish
bpinto: This is kinda unusual as we are running tests from multiple plugins at the same time, so I understand why you decided to prepend the plugin name, but maybe this is something we need to handle later, specifically for this project. We will discuss this later :)

bpinto:
Are you sure about -E? I haven't found anything like that on the Internet and my centos machine does support -E.

Those options are cool, and will be important for the suite. So, are you ready? Please don't add more functionality, just fix what is broken.

plugins/fish-spec/spec.eval.fish
bpinto: This is causing many tests to fail.

plugins/fish-spec/spec/suite.spec.fish
bpinto: This should be used instead.

plugins/fish-spec/spec.run.fish
```diff
@@ -36,27 +50,44 @@
...
+    set -l failed 0
+
+    for test in $tests
+      spec.eval before_each $argv
+
+      # Capture error dump coming from stdout in case test fails.
+      if not set -l stdout (spec.eval $test --tab 1 $argv)
+        set result 1 # Flunk!
+
+        test -n "$stdout"
```
bpinto: `if` instead.

bucaran:
Done. Tests are passing.

I went back to using stdout and with almost NO hassle the tests are now passing, and output is looking pretty neat.

Now, I'll look into using signals to see if that would actually improve anything, but at a glance you can see things are simpler now, it starting to look like an MVC app.

```fish
spec.run                # model 
  spec.eval test      # controller
  ... 
  spec.view output # view
```

Just take a look at `spec.eval` (controller)

```fish
import plugins/msg

function spec.eval
  set -l result $status
  set -l test $argv[1]
  set -e argv[1]

  if functions -q $test
    # Run the test yielding control to the user defined spec.fish test.
    set -l output (eval $test)
    set result $status
    # Display test results.
    spec.view $test $result $argv -- $output
  end
  return $result
end
```

Ready to push...

bucaran:
Check this out!
mov

bpinto:
It looks very nice.

Tests are failing on CI though.

```
spec.run                # model 
  spec.eval test      # controller
  ... 
  spec.view output # view
```

Funny abstraction because model is talking to the controller and the view, which is unusual in a MVC pattern. :)

bucaran:
> Are you sure about -E? I haven't found anything like that on the Internet and my centos machine does support -E.

Maybe you have the GNU `grep` in OS X.

This is about [`sed`](http://unix.stackexchange.com/questions/13711/differences-between-sed-on-mac-osx-and-other-standard-sed), but the same applies to other utilities.

bpinto:
Well, then it seems to be fine for grep, as my centos machine supports it.

plugins/bak/spec/bak.spec.fish
```diff
@@ -1,7 +1,7 @@
 import plugins/bak
 import plugins/fish-spec
 
-function describe_bak -d 'Testing bak plugin'
+function describe_plugins:bak -d 'bak plugin'
```
bpinto: I would keep `describe_bak`. Most of the tests will belong to a plugin, we don't need to enforce people to have a similar function name.

plugins/fish-spec/expect.fish
bpinto: Have you read this comment?

bucaran:
It is well known there are compatibility issues between OS X BSD version of GNU tools in Linux, etc. But if it works for you, then yippee!

plugins/bak/spec/bak.spec.fish
bucaran: Ah, that. Agreed. It's just a name.

plugins/fish-spec/expect.fish
```diff
@@ -51,6 +60,11 @@ function expect
     case --to-be-true
       eval "$expected"
       test $status -eq 0
+    case --to-contain-any
```
bpinto: This is not documented, nor tested. I'd remove this and add in a future PR.

bucaran:
```
spec.run                # controller
  spec.eval test      # model
  ... 
  spec.view output # view
```
Maybe?

Actually, I think spec.run and spec.eval are both types of controllers.

plugins/fish-spec/expect.fish
bucaran: Yup.

plugins/fish-spec/spec.eval.fish
```diff
@@ -1,98 +1,29 @@
...
-        end
-      case -s --silent
-        set silent $argv[$index]
-      case -u --unit
-        # Evaluate query as a single function instead of a set.
-        set tests $query
-    end
-  end
+function spec.eval
+  set -l result $status
```
bpinto: Why do we set result using `$status` if we haven't evaluated anything yet?

plugins/fish-spec/expect.fish
bucaran: Got it. Comments gotta go.

plugins/fish-spec/spec.run.fish
```diff
@@ -36,27 +50,44 @@
...
-  spec.eval describe_ --depth 0 $output
-  spec.eval before_all $output
-  for test in (spec.functions "it_")
-    spec.eval before_each $output
-
-    # Flunk test if any single test fails, but do not stop the suite.
-    if not spec.eval $test --unit --depth 1 $output
-      set result 1
+  set -l result  0
+  set -l tests  ""
```
bpinto: I think this `set` is not necessary.

plugins/fish-spec/spec.eval.fish
bucaran: Because if the function does not exist (i.e., it is not implemented in the suite `before_each`, etc.) the `if functions -q ...` will change the value of `$status`, causing havoc and disaster.

plugins/fish-spec/spec/expect.to_be_true.spec.fish
bucaran: :+1:

plugins/fish-spec/spec.eval.fish
bpinto: Shouldn't we set either `0` or `1` specifically?

plugins/fish-spec/spec/list.erase.spec.fish
bpinto: ping @bucaran 

```
function it_returns_0_if_items_are_erased_from_some_but_not_all_lists
  list.erase 1 2 3 --from nums_until_10 odds_until_10
  expect $status --to-equal 0
end
```

spec/oh-my-fish.spec.fish
```diff
@@ -13,25 +13,20 @@ function describe_oh_my_fish -d "Oh-My-Fish test spec..."
...
   end
 
-  function it_has_a_default_custom_directory \
-    -d "It should have a default custom directory"
-
+  function it_has_a_default_custom_directory
     set -e fish_custom
     load_oh_my_fish
     expect $fish_custom --to-eq "$HOME/.oh-my-fish/custom"
+    expect $fish_custom --to-eq "$HOME/.oh-my-fish/custom"
```
bpinto: Double expect.

plugins/fish-spec/spec/spec.spec.fish
```diff
@@ -0,0 +1,130 @@
...
+    set -l suite "
+        import plugins/fish-spec
+        function describe_blank_suite
+        end
+        spec.run
+      "
+    expect (run_nested_suite $suite) --to-equal "No tests found."
+  end
+
+  function it_runs_all_describe_blocks
```
bpinto: Why is this so complicated now?

```
    set -l suite "
        import plugins/fish-spec
        function describe_blank_suite
          echo 'first describe'
        end
        function describe_another_blank_suite
          echo 'second describe'
        end
        spec.run
      "

    expect (run_nested_suite $suite) --to-contain 'first describe' 'second describe'
```

This is so much easier to understand.

plugins/fish-spec/spec/spec.spec.fish
```diff
@@ -0,0 +1,130 @@
...
+
+          function it_is_also_executed
+            return 1
+          end
+        end
+
+        spec.run
+      "
+    run_nested_suite $suite
+    expect $status --to-equal 1
```
bpinto: I don't think this guarantees that the tests were executed, the previous test was confirming that the tests were evaluated.

plugins/fish-spec/spec.eval.fish
bucaran: Let me explain this better, the buggy eval will cause this to fail, because an empty function (at least when the terminal is interactive will return 1).

Now, with the patched eval we could set this to 0, 1 or 42, it wouldn't matter, because the `$status` is always saved (inside eval) and reset just before the expression is evaluated. So, if the expression does not change `$status` (e.g., an empty function, or a function that just uses `set`, etc.) the `$status` will be set to whatever it was before.

Bottom line, we don't have to initialize this variable to `$status`. But we might as well just do that to implicate we don't care about `$status` anyway. Initializing it to 0 or 1 would be beg the question however.

spec/oh-my-fish.spec.fish
bucaran: Thanks

plugins/fish-spec/spec/spec.spec.fish
bucaran: Why? `$status` will be zero if the last test does not run.

plugins/fish-spec/spec.eval.fish
bpinto:
```
λ eval
echo $status
0
```

Eval returns 0 if there is nothing to eval, so if there is no test to eval, we could follow eval's convention and also return 0, meaning we could initialize this variable with 0.

plugins/fish-spec/spec/spec.spec.fish
bucaran: I went over the top with `it_runs_all_describe_blocks`. I could have made it simpler without having to create several suites. Let's work on these tests later.

plugins/fish-spec/spec/spec.spec.fish
bpinto: This is testing the framework, we could have a bug on the framework where we always return 1 without evaluating the tests, for instance. Or we could only run the last test, without running the two previous.

There are many scenarios where this could give a false positive.

plugins/fish-spec/spec/spec.spec.fish
bpinto: Let's work on it. I notice I couldn't produce an output, this is something I didn't want to raise before we got this branch ready to merge, which I think we are now.

But not being able to output is bad, because I can no longer add debugging points to my test. But I'll work on converting this to using signals so we do not need to capture output.

plugins/fish-spec/spec/spec.spec.fish
bpinto: Let's work on this later.

plugins/fish-spec/spec/list.erase.spec.fish
bucaran: :+1:

plugins/fish-spec/spec.eval.fish
bucaran: Like I said, you can initialize it with any value you want. It will not do anything.

plugins/fish-spec/spec.run.fish
bucaran: It is important to declare it outside the `for` so that we can test later when checking if it's empty in order to say "No tests found."

plugins/fish-spec/spec.eval.fish
bpinto: If you send no arguments to `spec.eval` it will return 0 if we set it to 0 here.

plugins/fish-spec/spec/spec.spec.fish
bucaran: Fair enough. I'll be glad to help you with that too. I want to see what you come up with too! :+1:

bpinto: hehehhe your help will be more than appreciated! :)

plugins/fish-spec/spec/spec.spec.fish
bucaran: You are the tests man. Sure. I will add a new PR and let's merge with that.

plugins/fish-spec/spec/spec.spec.fish
bucaran: :feelsgood: Great.

plugins/fish-spec/spec/spec.spec.fish
bpinto: :feelsgood:

PR #321, opened 2015-01-16
added osx plugin

CYKEB:
I created a plugin with integration for osx, based on a similar plugin for oh-my-zsh


plugins/osx/osx.load
```diff
@@ -0,0 +1,104 @@
...
+#          FILE:  osx.fish
+#   DESCRIPTION:  oh-my-fish plugin file.
+#        AUTHOR:  Felix Sonntag
+#       VERSION:  1.0
+#       COMMENT:  Inspired by the oh-my-zsh plugin by Sorin Ionescu (sorin.ionescu@gmail.com)
+#                 https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/osx/osx.plugin.zsh
+# ------------------------------------------------------------------------------
+
+
+function tab
```
bpinto: Functions could receive a `-d` (description) argument so it's easier to understand what they do.
Each of the functions should live in their own file to make use of fish autoload functionality.

plugins/osx/osx.load
```diff
@@ -0,0 +1,104 @@
...
+  man -t $argv | open -f -a Preview
+end
+
+function trash
+  set -l trash_dir "$HOME/.Trash"
+  for item in $argv
+    if test -e $item
+      set -l item_name (basename $item)
+      if test -e "$trash_dir/$item_name"
+        set -l currentTime (date "+%H.%M.%S")
```
bpinto: Your code is snake cased, but `currentTime` is not.

plugins/osx/osx.load
```diff
@@ -0,0 +1,104 @@
...
+      else
+        mv -f "$item" "$trash_dir/"
+      end
+    end
+  end
+end
+
+function pfs
+  osascript 2>/dev/null -e '
+      set output to ""
```
bpinto: I don't understand this variable `output`, where is it used?

plugins/osx/osx.load
```diff
@@ -0,0 +1,104 @@
...
+        set output to output & ((item item_index of the_selection as alias)\'s POSIX path) & the_delimiter
+      end repeat'
+end
+
+
+function itunes
+  set -l opt $argv[1]
+  switch $opt
+    case launch play pause stop rewind resume quit
+      ;;
```
bpinto: Why do we need two `;`?

plugins/osx/osx.load
```diff
@@ -0,0 +1,104 @@
...
+        if item_index is item_count then set the_delimiter to ""
+        set output to output & ((item item_index of the_selection as alias)\'s POSIX path) & the_delimiter
+      end repeat'
+end
+
+
+function itunes
+  set -l opt $argv[1]
+  switch $opt
+    case launch play pause stop rewind resume quit
```
bpinto: Nothing necessary to send to itunes when one for any of these actions?

plugins/osx/osx.load
```diff
@@ -0,0 +1,104 @@
...
+      set opt "set mute to false"
+      ;;
+    case next previous
+      set opt "$opt track"
+      ;;
+    case vol volume
+      set opt "set sound volume to $argv[2]"
+      ;;
+    case "" -h --help
+      return 0
```
bpinto: Will this print the help below?

CYKEB: no, my fault

plugins/osx/osx.load
CYKEB: I was not sure about the code style, at least in bash this is used as an explicit break. Gets fixed

plugins/osx/osx.load
CYKEB: What do you mean with that?

bpinto: There is no `set opt "something"`

plugins/osx/osx.load
CYKEB: line 66

bpinto: Isn't this variable always `""`?

`set output to ""`?

plugins/osx/osx.load
CYKEB: `opt` is already set in line 72 and these commands don't take any additional options.

plugins/osx/osx.load
CYKEB: That's AppleScript code, so you don't reference a variable with a `$`

plugins/osx/osx.load
bucaran: There is no fall through inside `switch` statements in fish, so `break` is not required. 

`break` is only used inside loops, same as `continue`/

plugins/osx/osx.load
bpinto: :+1:

plugins/osx/OSX-README.markdown
```diff
@@ -0,0 +1,17 @@
+osx
```
bpinto: this file needs to be named README only otherwise github won't parse it.

plugins/osx/cdf.fish
```diff
@@ -0,0 +1,12 @@
+# cd to the current Finder directory
+
+function cdf
```
bpinto: `function cdf -d 'cd to the current Finder directory'`. This way fish shell will parse it and show it like this:

image

plugins/osx/itunes.fish
```diff
@@ -0,0 +1,36 @@
+# Play, pause etc. iTunes
+
+function itunes
+  if count $argv >/dev/null
```
bpinto: `if [ (count $argv) -gt 0 ]`?

plugins/osx/manp.fish
```diff
@@ -0,0 +1,12 @@
+# Open a specified man page in Preview
+
+function manp
+  if count $argv >/dev/null
+    switch $argv[1]
+      case -d --description
+        echo "Open a specified man page in Preview"
+        return 0
+    end
```
bpinto: No else handling no arguments.

plugins/osx/pfd.fish
```diff
@@ -0,0 +1,15 @@
+# Return the path of the frontmost Finder window
+
+function pfd
+  if count $argv >/dev/null
+    switch $argv[1]
+      case -d --description
+        echo "Return the path of the frontmost Finder window"
+        return 0
+    end
```
bpinto: No else handling no arguments.

plugins/osx/pfd.fish
```diff
@@ -0,0 +1,15 @@
+# Return the path of the frontmost Finder window
+
+function pfd
+  if count $argv >/dev/null
+    switch $argv[1]
+      case -d --description
```
bpinto: This is probably not necessary if the function had the description defined on its declaration.

CYKEB:
Thanks for all the feedback! I'm new to fish, so I'm sorry for the basic mistakes I have done.

CYKEB:
That's odd that the tests failed on linux. The plugin shouldn't affect these tests.

bpinto:
Don't worry about that, we all learn together :)

This test is randomly failing, but master has a possible fix, could you rebase your branch?

bpinto:
The changes look good, after the rebase we can merge it.

I have a custom OSX plugin already, where I keep some functions I wrote, guess I'll have to rename it. :)

CYKEB:
Great, thanks!
If your functions might be useful to other people you can extend the plugin :)

bpinto:
Thank @CYKEB

bpinto:
My functions were not osx related to be honest, I just didn't have a good name for it. :)
They are on my [dotfiles](https://github.com/bpinto/dotfiles), but they are not good enough to become plugins.

PR #324, opened 2015-01-17
run tests from correct repo on travis

bpinto:


bpinto:
@bucaran that branch tests should be green now.

PR #325, opened 2015-01-17
installing latest stable release and git built fish versions

bpinto:


.travis.yml
```diff
@@ -2,6 +2,7 @@ language: c
 
 env:
   - FISH_PPA=nightly-master
+  - FISH_PPA=release-2
```
bucaran: I originally had this too, but it failed like 70 times, and around that time I gave up, but I knew it would come back to hunt me! :laughing:

bpinto: lol, it is failing :P

bpinto:
Can I do this?

```
if status --is-interactive
  exec fish < /dev/tty
end

exit 0
```

Or will the `exit` cause the `fish` call to exit too?

bpinto:
Maybe `--is-interactive` is not available?

bucaran:
`exec fish < /dev/tty` get skipped in the travis build, that's why it's there.

PR #331, opened 2015-01-20
fish getopts plugin

bucaran:
## Tests

Tests are implemented via [Oh-My-Fish](https://github.com/bpinto/oh-my-fish) spec runner _fish-spec_.

## Premise
Every utility / function needs to handle arguments. What usually happens is that every function does all the grunt work to parse `$argv` itself, and while some patterns do tend to recur, it's different almost every time. Enter fish `getopts`.

## Synopsis
  `getopts <option string> [ARGV...]`

#### `<option string> [:][<letter>:][<word>[:[^]]]`

A string containing the option characters recognized by the utility calling `getopts`. If a `<letter>` or `<word>` ends in `:`, the option is expected to have an argument, which may be supplied separately or next to the option without spaces in the same string.

To indicate long options: `<letter>:<word>` and `<word>` are both valid option strings that will attempt to match `-<letter>` and `--<word>`.

To indicate optional arguments, use a `^` character after a `:` at the end of the option in the option string. For example `<letter>:^` and `<letter>:<word>:^` are both valid. Optional arguments should be supplied in the same string as the option and without spaces, e.g, `-<letter>value` will correctly assign `value` as the argument to the the option `<letter>`, but `-<letter>` value, will parse `value` as the next argument in `<ARGV>`.

To specify optional arguments using the option's long form, use a `=` character after the option: `--<word>=value`.

Use a `:` at the beginning of the option string to enable strict mode. If enabled, `getopts` will exit with a status __> 0__ if or when an unknown option is found. See _DIAGNOSTICS_.

#### `<ARGV>`

List of options and operands to parse. `getopts` prints any matched options as well as available argument separated by a \n to stdout and returns with a status of __0__ if there are still arguments; else returns with a status __> 0__ if the end of the options is reached or an error occurs. See _DIAGNOSTICS_.

## Description
`getopts` obtains options and their arguments from a list of parameters that, as indicated by <option string>, are single letters preceded by a `-` or words preceded by `--` and possibly followed by an argument value.

fish `getopts` follows the specifications described in the [Utility Syntax Guidelines](http://pubs.opengroup.org/onlinepubs/7908799/xbd/utilconv.html). The following is a summary of the features:

+ Short options; single letters preceded by `-`, and long options; words preceded by `--`, are both supported.

+ Single letters may be grouped. `-abc` → `-a -b -c`

+ Options required to take an argument can specify the argument either in the same string as the option or separated from the by a space. (1) `-a argument`, (2) `-aargument`

+ Options that can take an argument optionally shall specify the argument in the same string as the option argument if in short option style: `-aargument`, or separated by a `=` if in long form: `--long-form=argument`. If a blank space is used, the following argument will be treated independently.

+ Options can appear multiple times in the same argument list. `getopts` will print every match sequentally on each call, and should default to the short form of the option if available.

+ The option delimiter `:` and optional argument character `^` shall not be used as an option.

+ `getopts` will return the remaining operands when the end of the options is reached, i.e, a `--` argument that is not an option is found, or an argument that does not begin with `-` is found.

## Environment Variables
The following environment variables are used internally by `getopts`. These variables are erased from memory when the functions returns 1.

+ `__getopts_optstr` _Whitespace trimmed option string._

+ `__getopts_argv` _Preprocessed copy of arguments._

+ `__getopts_index` _Index of the next argument to handle._

+ `__getopts_required` _List of options with required arguments._

+ `__getopts_optional` _List of options with optional arguments._

## Diagnostics
Possible exit status values are:

+ __0__: An argument formed like an option was found. This causes getopts to print the option short-style and its argument if avaiable. If strict-mode is enabled setting the first character of the option to `:`, an unknown option will cause `getopts` to fail. See below.

+ __1__: The end of the options is reached. Also sends remaining operands to stdout.

+ __2__: An option argument was missing.

+ __3__: An unknown option was found. Only if strict-mode is enabled.

## Examples
```fish
  function my_utility
    while set optarg (getopts "l:long x: o:optional:^" $argv)
      switch $optarg[1]
        case l
          echo handle `-l --long`
        case x
          echo handle `-x` w/ argument `$optarg[2]`
        case o
          echo handle `-o --optional` w/ optional argument `$optarg[2]`
        case \*
          echo unknown option `$optarg[1]`
      end
    end
    echo -n operands: "`$optarg`"
  end
```

## Links
+ [UNIX Utility Conventions](http://pubs.opengroup.org/onlinepubs/7908799/xbd/utilconv.html)
   
## License
[MIT](http://opensource.org/licenses/MIT) © [Jorge Bucaran](http://bucaran.me)

### DIFF
--- a/plugins/getopts/getopts.fish
+++ b/plugins/getopts/getopts.fish
@@ -1,0 +1,399 @@
+# NAME
+#   getopts -- getopts for fish
+#
+# SYNOPSIS
+#   getopts <option string> [ARGV...]
+#
+# OPTIONS
+#   <option string> [:][<letter>:][<word>[:[^]]]
+#
+#   A string containing the option characters recognized by the utility
+#   calling getopts. If a <letter> or <word> ends in `:`, the option is
+#   expected to have an argument, which may be supplied separately or
+#   next to the option without spaces in the same string.
+#
+#   To indicate long options: <letter>:<word> and <word> are both valid
+#   option strings that will attempt to match -<letter> and --<word>.
+#
+#   To indicate optional arguments, use a `^` character after a `:` at
+#   the end of the option in the option string. For example <letter>:^
+#   and <letter>:<word>:^ are both valid. Optional arguments should be
+#   supplied in the same string as the option and without spaces, e.g,
+#   -<letter>value will correctly assign `value` as the argument to the
+#   the option <letter>, but -<letter> value, will parse `value` as the
+#   next argument in <argv>.
+#
+#   To specify optional arguments using the option's long form, use a
+#   `=` character after the option: --<word>=value.
+#
+#   Use a `:` at the beginning of the option string to enable strict
+#   mode. If enabled, getopts will exit with a status > 0 if or when
+#   an unknown option is found. See DIAGNOSTICS.
+#
+#   <ARGV>
+#
+#   List of options and operands to parse. getopts prints any matched
+#   options as well as available argument separated by a \n to stdout
+#   and returns with a status of 0 if there are still arguments; else
+#   returns with a status > 0 if the end of the options is reached or
+#   an error occurs. See DIAGNOSTICS.
+#
+# DESCRIPTION
+#   getopts obtains options and their arguments from a list of parameters
+#   that, as indicated by <option string>, are single letters preceded by
+#   a - or words preceded by -- and possibly followed by an argument value.
+#
+#   fish getopts follows the specifications described in the Utility Syntax
+#   Guidelines (see LINKS); the following is a summary of the features:
+#
+#     + Short options; single letters preceded by -, and long options;
+#       words preceded by --, are both supported.
+#
+#     + Single letters may be grouped. -abc → -a -b -c
+#
+#     + Options required to take an argument can specify the argument
+#       either in the same string as the option or separated from the
+#       by a space. (1) -a argument, (2) -aargument
+#
+#     + Options that can take an argument optionally shall specify the
+#       argument in the same string as the option argument if in short
+#       option style: -aargument, or separated by a = if in long form:
+#       --long-form=argument. If a blank space is used, the following
+#       argument will be treated independently.
+#
+#     + Options can appear multiple times in the same argument list.
+#       getopts will print every match sequentally on each call, and
+#       should default to the short form of the option if available.
+#
+#     + The option delimiter `:` and optional argument character `^`
+#       shall not be used as an option.
+#
+#     + getopts will return the remaining operands when the end of the
+#       options is reached, i.e, a `--` argument that is not an option
+#       is found, or an argument that does not begin with `-` is found.
+#
+# ENVIRONMENT VARIABLES
+#   The following environment variables are used internally by getopts.
+#   These variables are erased from memory when the functions returns 1.
+#
+#     + __getopts_optstr      Whitespace trimmed option string.
+#     + __getopts_argv        Preprocessed copy of arguments.
+#     + __getopts_index       Index of the next argument to handle.
+#     + __getopts_required    List of options with required arguments.
+#     + __getopts_optional    List of options with optional arguments.
+#
+# DIAGNOSTICS
+#   Possible exit status values are:
+#
+#     0  An argument formed like an option was found. This causes getopts
+#        to print the option short-style and its argument if avaiable. If
+#        strict-mode is enabled setting the first character of the option
+#        to `:`, an unknown option will cause getopts to fail. See below.
+#
+#     1  The end of the options is reached. Also sends remaining operands
+#        to stdout.
+#
+#     2  An option argument was missing.
+#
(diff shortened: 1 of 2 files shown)


bucaran:
Please review and let me know if the spec looks good.

bucaran:
`bak` tests fails and either that breaks something causing everything else to fail, or I am missing something else. Of course, tests pass smooth here. Any ideas? What about disabling `bak` tests until there is a robust spec for it?

bpinto:
If the bak tests continue to fail, we should remove them, however unless I missed something, bak tests are not failing on the latest travis run.

```
.oh-my-fish:bucaran-getopts λ ./script/run-tests.fish -v
describe bak plugin
  ✔ it checks if bak filename pattern is followed
  ✔ it copies a single file
  ✔ it copies multiple files
  ✔ it moves a single file
  ✔ it moves multiple files
  ✔ it normalizes the file name
  ✔ it undo copies of a directory
  ✔ it undo copies of a single file
  ✔ it undo copies of multiple files
  ✔ it undo moves of a single file
  ✔ it undo moves of multiple files

...

  ✔ it runs all it blocks

8 passing

fish getopts
An error occurred while redirecting file descriptor 1
dup2: Bad file descriptor
An error occurred while redirecting file descriptor 1
dup2: Bad file descriptor
  ✔ it is loaded
printf: not enough arguments
  ✔ it reads a complex optstr
printf: not enough arguments
  ✔ it reads a long option with optional argument
  ✔ it reads a long option with optional argument missing
An error occurred while redirecting file descriptor 1
dup2: Bad file descriptor
  ✔ it reads a long style only option with argument
  ✔ it reads a long style option
  ✔ it reads a short style option
  ✔ it reads a single short option
  ✔ it reads a single short option with optional argument
  ✔ it reads a single short option with optional argument missing
  ✔ it reads an option with argument
  ✔ it reads an option with argument joined
  ✔ it reads multiple mixed style options with argument
  ✔ it reads multiple options with optional argument mixed
  ✔ it reads multiple short options
  ✔ it reads multiple short style options with argument
  ✔ it reads options with all optional args missing
printf: not enough arguments
  ✔ it returns 1 if empty
printf: not enough arguments

18 passing
```

bucaran:
I think the problem is that we are not using the travis environment variables inside the install script. `bootstrap.sh` is correctly setup to curl from the PR's author repository `tools/install.fish`, but the script itself will always download from `bpinto` which is not what we want when using a CI.

bpinto:
The code above was from running the tests locally, not copying from Travis.

bucaran:
What is exactly failing in your computer?

I see `18 passing`.

bpinto:
That is true, the end result was "18 passing" but many "dup2: Bad file descriptor" were printed. If this is expected, we need to silent the command or if it's not, fix it.

I suggest to also remove all bak plugin tests. I think the plugin needs a refactoring on the code anyway. :P

bucaran:
I digging in the travis logs to see if I can get any clues on the Bad file descriptor error.

bucaran:
Maybe the problem was `while getopts >/dev/null ^/dev/null; end`...

Which I do `after_each` test to make sure I exit `getopts` correctly before the next test..............

bucaran:
[Looking at the travis build for 145](https://travis-ci.org/bpinto/oh-my-fish/builds/47682700)
```
$ git clone --depth=50 git://github.com/bpinto/oh-my-fish.git bpinto/oh-my-fish
Cloning into 'bpinto/oh-my-fish'...
```
Why?

bucaran:
I tried changing the plugin name thinking it was conflicting with an existing `getopts`, but it still says the command is not found. It is pulling from your repo as you can see in the travis logs.

`$ git clone --depth=50 git://github.com/bpinto/oh-my-fish.git bpinto/oh-my-fish`

bpinto:
I don't doubt that there could be something wrong with the installation.

It's just weird that the test is pulled but the code is not.

bucaran:
Mmmm. Do you mind trying the tests again locally? If it runs well, we could merge, and then travis would probably work again.

bpinto:
This is the output now:

```
fish getopts
  ✔ it reads a complex optstr
  ✔ it reads a long option with optional argument
  ✔ it reads a long option with optional argument missing
  ✔ it reads a long style only option with argument
  ✔ it reads a long style option
  ✔ it reads a short style option
  ✔ it reads a single short option
  ✔ it reads a single short option with optional argument
  ✔ it reads a single short option with optional argument missing
  ✔ it reads an option with argument
  ✔ it reads an option with argument joined
  ✔ it reads multiple mixed style options with argument
  ✔ it reads multiple options with optional argument mixed
  ✔ it reads multiple short options
  ✔ it reads multiple short style options with argument
  ✔ it reads options with all optional args missing
printf: not enough arguments
  ✔ it returns 1 if empty
```

bpinto:
```
.oh-my-fish:bucaran-getopts λ which getopts
/usr/bin/getopts
```

We need to install `getopts` on .travis.yml file.

PR #333, opened 2015-01-20
add OSX emoji workaround to README

kholbekj:
Spend way too much time figuring out that the strange colored :anchor: didn't come from my font, but rather from the OSX emoji font.

Dunno if this is appropriate here, but it would have saved me a lot of time.

### DIFF
--- a/themes/ocean/README.md
+++ b/themes/ocean/README.md
@@ -24,3 +24,6 @@ This theme is based loosely on [agnoster](https://gist.github.com/agnoster/37128
     set -g theme display_user yes
     set -g default_user your_default_username
     ```
+
+#### Troubleshooting in iTerm
+If you see a colored anchor icon instead of the white one, this is because your system font is lacking the anchor icon, and iTerm defaults to the Apple Emoji Font. To get the right icon, either add it to your current font, or chose a powerline font that has it (like Meslo for Powerline, available in https://github.com/powerline/fonts), and select it for your Non-ASCII font.


bpinto:
Isn't there another solution as changing the character code on OSX?

kholbekj:
It's very possible, but so far I've not been able to figure a way to either change the font priority, or somehow tell OSX to not use the emoji font in a secific context.

I've googled like a maniac, I really have.

bobthecow:
at that point, wouldn't it make more sense to tell the user how to switch the glyph to something different, rather than giving instructions for a good way to globally break perfectly valid character support on their system?

kholbekj:
Sure, but how do you do that in a way that aligns to the terminal font?
I honestly have no idea.

kholbekj:
Maybe I could change it to something like `mv whatsthefile.tff whatsthefile.tff.old` for now? It's not that I'm eager or in a haste or anything, but I don't really know where to go from here, and I'd argue this information is still useful, regardless of breaking an annoying functionality.

bpinto:
image

I have this font and I believe the anchor is correct on my terminal. If this is the case, perhaps you need to configure your terminal to use another font?

kholbekj:
Your anchor looks like the emoji anchor I'm trying to get out of.
1__fish___users_kasper_src_kholbekj_workouter__fish_
That's what it's supposed to look like.

bpinto:
This is though... I'm not sure what could happen if someone deletes an apple font from their computer.

This is a quote for OSX 10.7:
> While you may never use it, Apple Color Emoji.ttf must be left in the System's Fonts folder. It will only cause a problem if you open the Character palette and click on the Emoji heading, but then the Character palette will be royally stuck.

From: http://www.jklstudios.com/misc/osxfonts.html

P.S.: Terminal correctly displays the unicode, maybe it's an [iterm bug](http://superuser.com/questions/477608/fancy-colored-rendering-issue-on-osx-10-8-retina-with-iterm)?

kholbekj:
What bothers me about that is it was supposedly 'fixed' in a beta from dec 2014.
I'll check out if I can find something then.

kholbekj:
Hm, it seems that quote mostly applies to Lion, at least my character palette is working, if a bit weird without the font: 
characters
No violent explosions though.

kholbekj:
I've made an Issue with iTerm https://code.google.com/p/iterm2/issues/detail?id=3384&thanks=3384&ts=1421825471

But I'll keep looking for better workarounds.

bpinto:
Thanks for the issue. I'm not sure about what to do here, I feel like removing the font is not the solution for this problem. What do you think?

kholbekj:
I get what you mean. It feels like an iTerm problem, but i also bugs me that it by default doesn't work in iTerm. 

But maybe putting this in the readme is too offensive, I guess not everyone considers the Apple Emoji Font a soulless annoyance that's taking valueble space from a used-to-be useful character palette.

bpinto:
:+1: I like the _soulless_ addition.

kholbekj:
Ouh hang on think i found something! https://code.google.com/p/iterm2/issues/detail?id=2062#c5

kholbekj:
Okay, I think there might be hope - it seems this comes from the glyph not being available in any powerline font i can find, defaulting to an OSX interpretation, which is inane enough prioritize Emoji. But I'll have to get hold of the Emoji font again to test... :+1:

bpinto:
Okay, I think I could work by following that issue (changing non-ascii font to Apple Symbols).

screen recording 2015-01-21 at 09-05

kholbekj:
That's unfortunately not a solution for this theme, as you'd need an 'apple symbols' for powerline fonts.
It'll break the prompt, and rather horribly. I think the correct workaround might be adding the anchor font to the font you're using, or maybe pointing to a compatible font if one exists.

bpinto:
lol, this is getting complicated.

kholbekj:
Yeah, in the ideal world you could scecify iterms fallback font prioritization, but i think the cooler thing to do here is try to find a compatible font. I might even find the time to add a bunch of glyphs to Inconsolata powerline and share that.

bpinto:
Keep us posted.

kholbekj:
Found one! Meslo for powerline has the nessecary components! Changing the readme

kholbekj:
@bpinto Can you check if you like this explanation? :)

bpinto:
Perfect!

PR #334, opened 2015-01-20
install should clone from specified repo slug when building via a CI system

bucaran:
We were not using the travis environment variables inside the install script. `bootstrap.sh` is correctly setup to curl from the PR's author repository `tools/install.fish`, but the script itself will always download from `bpinto` which is not what we want when using a CI.

### DIFF
--- a/script/bootstrap.sh
+++ b/script/bootstrap.sh
@@ -9,4 +9,4 @@ else
   brew install fish $BREW_OPTIONS
 fi
 
-curl -L https://github.com/$TRAVIS_REPO_SLUG/raw/$TRAVIS_BRANCH/tools/install.fish | fish
+curl -L https://github.com/$TRAVIS_REPO_SLUG/raw/$TRAVIS_BRANCH/tools/install.fish | REPO_SLUG=$TRAVIS_REPO_SLUG fish
--- a/tools/install.fish
+++ b/tools/install.fish
@@ -16,8 +16,10 @@
 # The installation process is split up in the following steps:
 #
 #   1. Resolve source repository to default or get from environment variable.
-#      This allows installers curl -L ..install.fish | TRAVIS_REPO_SLUG=<fork> fish in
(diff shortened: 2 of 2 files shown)


bpinto:
I have tested this before and it was doing the correct thing. Can you push some commits before these changes and output the git remote and git branch name? I did this test before and they had the correct output.

bucaran:
You did setup `bootstrap.sh` correctly, but `install.fish` was still pulling from `bpinto` because no environment variables were being passed into the `fish` pipe that receives `install.fish` via curl.

I am testing `getopts` in another branch, `improve-install+getopts`. If it works it should prove my point.

bucaran:
> Can you push some commits before these changes and output the git remote and git branch name?

Is not necessary, we can look at the travis log.

bpinto:
What did you see at the travis log?

bucaran:
This is the current code:

```fish
# Allow installers to specify the source repository.
if not set -q TRAVIS_REPO_SLUG
  set TRAVIS_REPO_SLUG bpinto/oh-my-fish
end

# Abort installation if oh-my-fish is already installed.
if test -d $fish_path
  log yellow "You already have Oh My Fish installed."
  log white  "Remove $fish_path if you want to reinstall it."
  exit 1
end
```

How are we pulling from the PR's author branch?

bpinto:
Travis sets the environment variable $TRAVIS_REPO_SLUG and we are reusing it on this script.

bucaran:
That's what I thought, but I can't see where it gets passed to the script in

bpinto:
It's on the computer, it's exported, so the script has access to it.
If we were not running the tests on your branch, we wouldn't have your tests running.

bucaran:
> It's on the computer, it's exported, so the script has access to it.
Gotcha.

Mmmm, so why do my tests fail if they do pass here?

bpinto:
They also fail on my computer. So, we need to investigate the issue.

PR #337, opened 2015-01-20
Fix some issues with the usage of RBENV_ROOT.

daenney:
`RBENV_ROOT` can be used to relocate where rbenv ends up storing the shims
and versions of Ruby it installs. Because of this split `RBENV_ROOT`
cannot be used to reliably construct a path to the rbenv binary itself.

Instead we now rely on `$RBENV_BIN_ROOT` pointing to the `bin/` directory
containing the rbenv binary/script.

If `$RBENV_BIN_ROOT` is not set we search `$PATH` first and then the
user's home directory making sure to prepend the necessary directories to
our `$PATH` depending on where we find rbenv.

Lastly we prepend to our `$PATH` the location of `(rbenv root)/shims` so
that we can find the shims rbenv generates for us.

### DIFF
--- a/plugins/rbenv/rbenv.load
+++ b/plugins/rbenv/rbenv.load
@@ -1,15 +1,28 @@
-set -l rbenv_dir "$RBENV_ROOT"
-if [ ! $rbenv_dir ]
-  set rbenv_dir $HOME/.rbenv
+# Try to find rbenv by:
+# * Looking for $RBENV_BIN_ROOT/rbenv. This environment variable is specific to this plugin
+# * Looking for rbenv on $PATH
+# * Looking for rbenv in the user's home directory
+# When necessary, prepend the bin/ directory containing rbenv to our $PATH.
+if [ -e "$RBENV_BIN_ROOT/rbenv" ]
+  set rbenv_binary "$RBENV_BIN_ROOT/rbenv"
+  _prepend_path "$RBENV_BIN_ROOT"
+else if [ (command command -v rbenv) ]
+  set rbenv_binary (command command -v rbenv)
+else if [ -e "$HOME/.rbenv/bin/rbenv" ]
+  set rbenv_binary "$HOME/.rbenv/bin/rbenv"
+  _prepend_path "$HOME/.rbenv/bin"
+else
+  echo "Could not find rbenv. Make sure it's on your system path, in your home directory or set the RBENV_BIN_ROOT environment variable pointing to the directory where you unpacked rbenv."
+  exit 1
 end
 
 set -l supports_fish
 set -l supports_fish_version '0.4.0'
-set -l user_version (eval $rbenv_dir/bin/rbenv --version | sed -E 's/^rbenv ([[:digit:]\.]{2,}).*$/\1/g')
+set -l user_version (eval $rbenv_binary --version | sed -E 's/^rbenv ([[:digit:]\.]{2,}).*$/\1/g')
 
 if [ $user_version = $supports_fish_version ]
   set -l supports_fish_commits '56'
-  set -l user_commits (eval $rbenv_dir/bin/rbenv --version | sed -E 's/^.+-([[:digit:]]{1,}).+$/\1/g')
+  set -l user_commits (eval $rbenv_binary --version | sed -E 's/^.+-([[:digit:]]{1,}).+$/\1/g')
   if [ $user_commits -ge $supports_fish_commits ]
     set supports_fish true
   end
@@ -25,9 +38,12 @@ else
   end
 end
 
-_prepend_path $rbenv_dir/bin
 if [ $supports_fish ]
-  status --is-interactive; and source (eval $rbenv_dir/bin/rbenv init - | psub)
+  status --is-interactive; and source (eval $rbenv_binary init - | psub)
 else
-  _prepend_path $rbenv_dir/shims
+  if [ "$RBENV_ROOT" ]
+    _prepend_path "$RBENV_ROOT/shims"
+  else
+    _prepend_path "$HOME/.rbenv/shims"
+  end
 end


daenney:
One thing, I didn't know how to stop the plugin from further loading so right now there's only an `echo` in there that goes "something is wrong". How do we gracefully handle this?

daenney:
I think there's a couple of mistakes in there. Need to look at it a bit more but this is the general idea.

bpinto:
If you just rename _RBENV_ROOT_ to _RBENV_BIN_ROOT_. Wouldn't it work?

To get the rbenv binary location we can use `which` instead:

```
λ which abcd
λ echo $status
1
λ which rbenv
/opt/boxen/rbenv/bin/rbenv
```

daenney:
Just renaming `RBENV_ROOT` to `RBENV_BIN_ROOT` doesn't solve the issue as in the current implementation that's still being used to build up a path to the binary.

I'm an idiot for not thinking of which, I'll fix that right up.

bpinto:
> Just renaming RBENV_ROOT to RBENV_BIN_ROOT doesn't solve anything. That environment variable is being read to later on construct the path to the binary from. Reading out RBENV_BIN_ROOT would simply cause us to read a non-existent variable. At least from the code I'm interpreting it as "assign the value of $RBENV_ROOT to our internal rbenv_bin_dir variable"

I didn't understand, but I'll wait for your next push to re-read it.

daenney:
This is a much cleaner and conciser implementation.

daenney:
The problem with `RBENV_ROOT` is that it's a user-settable environment variable that affects the behaviour of rbenv. It's used to split rbenv in two 'locations', one location where the rbenv installation itself is located and one where rbenv data is to be stored. If set, `RBENV_ROOT` is the parent of that data directory and usually contains `shims/` and `versions/`.

Therefor we cannot rely on reading the `RBENV_ROOT` environment variable and constructing the path to the actual rbenv binary that way.

bpinto:
Yep, you are correct.

This is why I'm suggesting us to rename the environment variable we were using inside the function to set the binary dir from _RBENV_ROOT_ to _RBENV_BIN_ROOT_ so I can continue using my rbenv from: `/opt/boxen/rbenv/bin/rbenv`.

So instead of `set -gx RBENV_ROOT /opt/boxen/rbenv/bin/rbenv`, I'll do `set -gx RBENV_BIN_ROOT /opt/boxen/rbenv/bin/rbenv`

daenney:
Hmm, for some reason the `set -l (command command -v rbenv)` output isn't being captured. If I execute the command I get what I expect but the variable ends up empty.

bpinto:
Why not use `which rbenv`?

daenney:
Ah, I wasn't aware people where using it for that too!

daenney:
As far as why I used `command`, it's the POSIX-compliant way of doing this. I'm just not understanding why the output is lost.

plugins/rbenv/rbenv.load
```diff
@@ -1,15 +1,23 @@
-set -l rbenv_dir "$RBENV_ROOT"
-if [ ! $rbenv_dir ]
-  set rbenv_dir $HOME/.rbenv
+# Use the built-in command to make sure we call the actuall command
+# utility to figure out if rbenv is on the path and where
+set -l $rbenv_binary (command command -v rbenv)
```
bpinto: rename `$rbenv_binary` to `rbenv_binary`

bpinto:
`command -v` :+1:

plugins/rbenv/rbenv.load
daenney: *facepalm*

daenney:
Would you be okay with `RBENV_BIN_ROOT` pointing to the `bin/` directory instead of the binary? It makes more sense considering the name and it's useful to have so we can directly call `_prepend_path` on it.

bpinto:
Yep!

daenney:
Uhm, wait, wat. How did this test fail?

bpinto:
I have removed all `bak` plugin tests, they are flaky. Rebasing should fix your broken test.

plugins/rbenv/rbenv.load
```diff
@@ -1,15 +1,28 @@
...
+if [ -e "$RBENV_BIN_ROOT/rbenv" ]
+  set rbenv_binary "$RBENV_BIN_ROOT/rbenv"
+  _prepend_path "$RBENV_BIN_ROOT"
+else if [ (command command -v rbenv) ]
+  set rbenv_binary (command command -v rbenv)
+else if [ -e "$HOME/.rbenv/bin/rbenv" ]
+  set rbenv_binary "$HOME/.rbenv/bin/rbenv"
+  _prepend_path "$HOME/.rbenv/bin"
+else
+  echo "Could not find rbenv. Make sure it's on your system path, in your home directory or set the RBENV_BIN_ROOT enironment variable pointing to the directory where you unpacked rbenv."
```
bpinto: Typo: eniroment.

plugins/rbenv/rbenv.load
```diff
@@ -25,9 +38,8 @@ else
   end
 end
 
-_prepend_path $rbenv_dir/bin
 if [ $supports_fish ]
-  status --is-interactive; and source (eval $rbenv_dir/bin/rbenv init - | psub)
+  status --is-interactive; and source (eval $rbenv_binary init - | psub)
 else
-  _prepend_path $rbenv_dir/shims
+  _prepend_path (eval $rbenv_binary root)/shims
```
bpinto: Can't we infer this based on the RBENV_BIN_ROOT, RBENV_ROOT and rbenv_binary variables? Eval is slow.

daenney: Mmm, yes we can. When `RBENV_ROOT` is not set it defaults to `$HOME/.rbenv` so depending on that we need to either `_prepend_path $RBENV_ROOT/shims` or `_prepend_path $HOME/.rbenv/shims`

bpinto: I think we should avoid `eval` at all cost. From what I read from rbenv's owner, he does not recommend RBENV_ROOT so I don't expect him to change this any time soon. And if he does, we can always add this `eval`. :)

daenney: Perfect. I'll update this in a sec.

bpinto:
@daenney u there? :)

daenney:
Oh sorry, slipped my mind.

bpinto:
Merged.

If useful for anyone, this is how my env vars look like now:

```
set -gx RBENV_ROOT $BOXEN_HOME/rbenv
set -gx RBENV_BIN_ROOT $RBENV_ROOT/bin
```

PR #339, opened 2015-01-21
~ fish getopts plugin ~ 

bucaran:
## Tests

Tests are implemented via [Oh-My-Fish](https://github.com/bpinto/oh-my-fish) spec runner _fish-spec_.

## Premise
Every utility / function needs to handle arguments. What usually happens is that every function does all the grunt work to parse `$argv` itself, and while some patterns do tend to recur, it's different almost every time. Enter fish `getopts`.

## Synopsis
  `getopts <option string> [ARGV...]`

#### `<option string> [:][<letter>:][<word>[:[^]]]`
A string containing the option characters recognized by the utility calling `getopts`. If a `<letter>` or `<word>` ends in `:`, the option is expected to have an argument, which may be supplied separately or next to the option without spaces in the same string.

To indicate long options: `<letter>:<word>` and `<word>` are both valid option strings that will attempt to match `-<letter>` and `--<word>`.

To indicate optional arguments, use a `^` character after a `:` at the end of the option in the option string. For example `<letter>:^` and `<letter>:<word>:^` are both valid. Optional arguments should be supplied in the same string as the option and without spaces, e.g, `-<letter>value` will correctly assign `value` as the argument to the the option `<letter>`, but `-<letter>` value, will parse `value` as the next argument in `<ARGV>`.

To specify optional arguments using the option's long form, use a `=` character after the option: `--<word>=value`.

Use a `:` at the beginning of the option string to enable strict mode. If enabled, `getopts` will exit with a status __> 0__ if or when an unknown option is found. See _DIAGNOSTICS_.

#### `<ARGV>`

List of options and operands to parse. `getopts` prints any matched options as well as available argument separated by a \n to stdout and returns with a status of __0__ if there are still arguments; else returns with a status __> 0__ if the end of the options is reached or an error occurs. See _DIAGNOSTICS_.

#### Summary
Consult the table below for a quick visual reference. 

| Option String          |  Argument     | Short Style     | Long Style   |   
|:---------------:|:--------------:|------------------|:--------------------:|
   `"q:qqq:"`          | Required   | `-q1` `-q 1`    | `--qqq 1`           | 
   `"q:qqq:^"`         | Optional    | `-q1`              | `--qqq=3`              |

## Description
`getopts` obtains options and their arguments from a list of parameters that, as indicated by <option string>, are single letters preceded by a `-` or words preceded by `--` and possibly followed by an argument value.

fish `getopts` follows the specifications described in the [Utility Syntax Guidelines](http://pubs.opengroup.org/onlinepubs/7908799/xbd/utilconv.html). The following is a summary of the features:

+ Short options; single letters preceded by `-`, and long options; words preceded by `--`, are both supported.

+ Single letters may be grouped. `-abc` → `-a -b -c`

+ Options required to take an argument can specify the argument either in the same string as the option or separated from the by a space. (1) `-a argument`, (2) `-aargument`

+ Options that can take an argument optionally shall specify the argument in the same string as the option argument if in short option style: `-aargument`, or separated by a `=` if in long form: `--long-form=argument`. If a blank space is used, the following argument will be treated independently.

+ Options can appear multiple times in the same argument list. `getopts` will print every match sequentally on each call, and should default to the short form of the option if available.

+ The option delimiter `:` and optional argument character `^` shall not be used as an option.

+ `getopts` will return the remaining operands when the end of the options is reached, i.e, a `--` argument that is not an option is found, or an argument that does not begin with `-` is found.

## Environment Variables
The following environment variables are used internally by `getopts`. These variables are erased from memory when the function returns 1.

+ `__getopts_optstr` _Whitespace trimmed option string._

+ `__getopts_argv` _Preprocessed copy of arguments._

+ `__getopts_index` _Index of the next argument to handle._

+ `__getopts_required` _List of options with required arguments._

+ `__getopts_optional` _List of options with optional arguments._

## Diagnostics
Possible exit status values are:

+ __0__: An argument formed like an option was found. This causes getopts to print the option short-style and its argument if avaiable. If strict-mode is enabled setting the first character of the option to `:`, an unknown option will cause `getopts` to fail. See below.

+ __1__: The end of the options is reached. Also sends remaining operands to stdout.

+ __2__: An option argument was missing.

+ __3__: An unknown option was found. Only if strict-mode is enabled.

## Examples
```fish
  function my_utility
    while set optarg (getopts "l:long x: o:optional:^" $argv)
      switch $optarg[1]
        case l
          echo handle `-l --long`
        case x
          echo handle `-x` w/ argument `$optarg[2]`
        case o
          echo handle `-o --optional` w/ optional argument `$optarg[2]`
        case \*
          echo unknown option `$optarg[1]`
      end
    end
    echo -n operands: "`$optarg`"
  end
```

## Links
+ [UNIX Utility Conventions](http://pubs.opengroup.org/onlinepubs/7908799/xbd/utilconv.html)
   
## License
[MIT](http://opensource.org/licenses/MIT) © [Jorge Bucaran](http://bucaran.me)

### DIFF
--- a/plugins/getopts/getopts.fish
+++ b/plugins/getopts/getopts.fish
@@ -1,0 +1,409 @@
+# NAME
+#   getopts -- getopts for fish
+#
+# SYNOPSIS
+#   getopts <option string> [ARGV...]
+#
+# OPTIONS
+#   <option string> [:][<letter>][[:]<word>[:[^]]]
+#
+#   A string containing the option characters recognized by the utility
+#   calling getopts. If a <letter> or <word> ends in `:`, the option is
+#   expected to have an argument, which may be supplied separately or
+#   next to the option without spaces in the same string.
+#
+#   To indicate long options: <letter>:<word> and <word> are both valid
+#   option strings that will attempt to match -<letter> and --<word>.
+#
+#   For only short options, do not specify a :<word> after <letter>.
+#   For example, `a b` will match `-a` and/or `-b`.
+#
+#   To indicate optional arguments, use a `^` character after a `:` at
+#   the end of the option in the option string. For example <letter>:^
+#   and <letter>:<word>:^ are both valid. Optional arguments should be
+#   supplied in the same string as the option and without spaces, e.g,
+#   -<letter>value will correctly assign `value` as the argument to the
+#   the option <letter>, but -<letter> value, will parse `value` as the
+#   next argument in <argv>.
+#
+#   To specify optional arguments using the option's long form, use a
+#   `=` character after the option: --<word>=value.
+#
+#   Use a `:` at the beginning of the option string to enable strict
+#   mode. If enabled, getopts will exit with a status > 0 if or when
+#   an unknown option is found. See DIAGNOSTICS.
+#
+#   <ARGV>
+#
+#   List of options and operands to parse. getopts prints any matched
+#   options as well as available argument separated by a \n to stdout
+#   and returns with a status of 0 if there are still arguments; else
+#   returns with a status > 0 if the end of the options is reached or
+#   an error occurs. See DIAGNOSTICS.
+#
+# DESCRIPTION
+#   getopts obtains options and their arguments from a list of parameters
+#   that, as indicated by <option string>, are single letters preceded by
+#   a - or words preceded by -- and possibly followed by an argument value.
+#
+#   fish getopts follows the specifications described in the Utility Syntax
+#   Guidelines (see LINKS); the following is a summary of the features:
+#
+#     + Short options; single letters preceded by -, and long options;
+#       words preceded by --, are both supported.
+#
+#     + Single letters may be grouped. -abc → -a -b -c
+#
+#     + Options required to take an argument can specify the argument
+#       either in the same string as the option or separated from the
+#       by a space. (1) -a argument, (2) -aargument
+#
+#     + Options that can take an argument optionally shall specify the
+#       argument in the same string as the option argument if in short
+#       option style: -aargument, or separated by a = if in long form:
+#       --long-form=argument. If a blank space is used, the following
+#       argument will be treated independently.
+#
+#     + Options can appear multiple times in the same argument list.
+#       getopts will print every match sequentally on each call, and
+#       should default to the short form of the option if available.
+#
+#     + The option delimiter `:` and optional argument character `^`
+#       shall not be used as an option.
+#
+#     + getopts will return the remaining operands when the end of the
+#       options is reached, i.e, a `--` argument that is not an option
+#       is found, or an argument that does not begin with `-` is found.
+#
+# ENVIRONMENT VARIABLES
+#   The following environment variables are used internally by getopts.
+#   These variables are erased from memory when the functions returns 1.
+#
+#     + __getopts_optstr      Whitespace trimmed option string.
+#     + __getopts_argv        Preprocessed copy of arguments.
+#     + __getopts_index       Index of the next argument to handle.
+#     + __getopts_required    List of options with required arguments.
+#     + __getopts_optional    List of options with optional arguments.
+#
+# DIAGNOSTICS
+#   Possible exit status values are:
+#
+#     0  An argument formed like an option was found. This causes getopts
+#        to print the option short-style and its argument if avaiable. If
+#        strict-mode is enabled setting the first character of the option
+#        to `:`, an unknown option will cause getopts to fail. See below.
+#
+#     1  The end of the options was reached. Remaining operands are also
+#        sent to stdout.
+#
+#     2  An option argument was missing.
+#
+#     3  An unknown option was found. Only if strict-mode is enabled.
+#
+# EXAMPLES
+#   function my_utility
+#     while set optarg (getopts "l:long x: o:optional:^" $argv)
+#       switch $optarg[1]
+#         case l
+#           echo handle `-l --long`
+#         case x
+#           echo handle `-x` w/ argument `$optarg[2]`
+#         case o
+#           echo handle `-o --optional` w/ optional argument `$optarg[2]`
+#         case \*
+#           echo unknown option `$optarg[1]`
+#       end
+#     end
+#     echo -n operands: "`$optarg`"
+#   end
+#
+# LINKS
+#   UNIX Utility Conventions
+#   → http://pubs.opengroup.org/onlinepubs/7908799/xbd/utilconv.html
+#
+# AUTHORS
+#   Jorge Bucaran <@bucaran>
+#/
+function getopts
+  # Currently supported return success/error conditions.
+  set -l __CONTINUE                   0
+  set -l __END_OF_OPTIONS             1
+  set -l __OPTION_ARGUMENT_EXPECTED   2
+  set -l __UNKNOWN_OPTION_FOUND       3
+
+  # Self-destroying global variable cleanup utility.
+  # Should be called before returning 1.
+  function __getopts_cleanup
+    set -e __getopts_optstr
+    set -e __getopts_argv
+    set -e __getopts_index
+    set -e __getopts_required
+    set -e __getopts_optional
+    set -e __getopts_strict_mode
+    functions -e __getopts_increase_index
+  end
+
+  function __getopts_increase_index
+    set __getopts_index (math $__getopts_index + 1)
+  end
+
+  # Options string pre-processing.
+  if not set -q __getopts_optstr
+    set -g __getopts_optstr $argv[1]
+    set -e argv[1]
+
+    # Trim option string and collect required / optional options.
+    if [ -n "$__getopts_optstr" ]
+      set __getopts_optstr (printf $__getopts_optstr | tr '[:space:]' \n)
+
+      # Setting the first token of the option string to `:` enables
+      # strict mode. This causes getopts to abort the process if an
+      # unknown option is found.
+      set -l first_token (printf $__getopts_optstr | head -c1)
+
+      if [ : = "$first_token" ]
+        set -g __getopts_strict_mode --true
+
+        # We can safely remove the `:` character now.
+        set __getopts_optstr[1] (printf $__getopts_optstr[1] | cut -c2-)
+      end
+
+      # Collect options with optional and required option-arguments.
+      function __getopts_collect_optargs
+        for string in (printf "%s\n" $__getopts_optstr)
+          set -l token (printf $string | tail -c1)
+          set -l split_string (printf "%s\n" $string | tr : \n)
+
+          # Erase last to make sure not to append `^` token.
+          switch $token
+            case :
+              set -g __getopts_required $__getopts_required $split_string
+              set -e __getopts_required[-1]
+            case \^
+              set -g __getopts_optional $__getopts_optional $split_string
+              set -e __getopts_optional[-1]
+          end
+        end
+        functions -e __getopts_collect_optargs
+      end
+
+      __getopts_collect_optargs
+    end
+  end
+
+  # Sanitize arguments. Break up flags: -abc → -a -b -c, but skip optional
+  # arguments. If -w is a flag that can take an optional argument, -abw123
+  # should be parsed to → -a -b -w 123.
+  if not set -q __getopts_argv
+    function __getopts_sanitize_argv
+      for token in $argv
+        switch $token
+          case --\* # Skip!
+          case -\*
+            # Split each token into single characters with `.`
+            for char in (printf $token | cut -c2- | grep --only-matching .)
+              # Do not split short option characters if this option can
+              # take optional required arguments.
+              if [ -z "$suspend_break" ]
+                printf "-%s\n" $char
+              else
+                printf "%s" $char
+              end
+
+              # Suspend option break-up if the current option is either
+              # a required or an optional option type.
+              contains -- $char $__getopts_required $__getopts_optional
+                and set suspend_break --true
+
+              set last_char $char
+            end
+
+            # Break options if we were in suspend mode and reset flag.
+            # This makes sure to add a blank space if no argument was
+            # specified for the optional option-argument, but not for
+            # the required option-argument.
+            if [ -n "$suspend_break" ]
+              not contains -- $last_char $__getopts_required
+                and printf "\n"
+              set suspend_break ""
+            end
+            continue
+        end
+
+        printf "%s\n" $token
+      end
+      functions -e __getopts_sanitize_argv
+    end
+    set -g __getopts_argv (__getopts_sanitize_argv $argv)
+  end
+
+  # Always use our preprocessed argument list.
+  set argv $__getopts_argv
+
+  if not set -q __getopts_index
+    set -g __getopts_index 1
+  end
+
+  # Handle next argument as indicated by the current index.
+  if set -q argv[$__getopts_index]
+    set -l option $argv[$__getopts_index]
+
+    # Potential matches for $option are: (1) end of arguments `--`,
+    # (2) long/short options --* or -* or (3) begin of operands, if
+    # (1) and (2) fail.
+    switch $option
+      # End of arguments. Assume everything from this point are operands.
+      case --
+        set -e argv[1..$__getopts_index]
+        printf "%s\n" $argv
+        __getopts_cleanup
+        return $__END_OF_OPTIONS
+
+      # Looks like we have a well-formed long/short option. Parse.
+      case --\* -\*
+        # It will be useful later to know if the current option is in long
+        # style. Notice we add a leading blank, because cut fails when the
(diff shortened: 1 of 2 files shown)


bucaran:
@bpinto Good green :smile:

plugins/getopts/getopts.fish
```diff
@@ -0,0 +1,399 @@
...
+#
+#   To indicate long options: <letter>:<word> and <word> are both valid
+#   option strings that will attempt to match -<letter> and --<word>.
+#
+#   To indicate optional arguments, use a `^` character after a `:` at
+#   the end of the option in the option string. For example <letter>:^
+#   and <letter>:<word>:^ are both valid. Optional arguments should be
+#   supplied in the same string as the option and without spaces, e.g,
+#   -<letter>value will correctly assign `value` as the argument to the
+#   the option <letter>, but -<letter> value, will parse `value` as the
```
bpinto: > but `-<letter>` value, will parse value as the next argument in <ARGV>.

Did you mean
> but `-<letter> value`, will parse value as the next argument in <ARGV>.
?

plugins/getopts/getopts.fish
```diff
@@ -0,0 +1,399 @@
...
+#   next to the option without spaces in the same string.
+#
+#   To indicate long options: <letter>:<word> and <word> are both valid
+#   option strings that will attempt to match -<letter> and --<word>.
+#
+#   To indicate optional arguments, use a `^` character after a `:` at
+#   the end of the option in the option string. For example <letter>:^
+#   and <letter>:<word>:^ are both valid. Optional arguments should be
+#   supplied in the same string as the option and without spaces, e.g,
+#   -<letter>value will correctly assign `value` as the argument to the
```
bpinto: Is this `-<letter>value` from original getopts? This is wayyyyyyyy to ugly. :P

plugins/getopts/getopts.fish
```diff
@@ -0,0 +1,399 @@
...
+#     + The option delimiter `:` and optional argument character `^`
+#       shall not be used as an option.
+#
+#     + getopts will return the remaining operands when the end of the
+#       options is reached, i.e, a `--` argument that is not an option
+#       is found, or an argument that does not begin with `-` is found.
+#
+# ENVIRONMENT VARIABLES
+#   The following environment variables are used internally by getopts.
+#   These variables are erased from memory when the functions returns 1.
```
bpinto: What if it returns 0?

plugins/getopts/getopts.fish
```diff
@@ -0,0 +1,399 @@
...
+  # Self-destroying global variable cleanup utility.
+  # Should always be called before returning 1.
+  function __getopts_cleanup
+    set -e __getopts_optstr
+    set -e __getopts_argv
+    set -e __getopts_index
+    set -e __getopts_required
+    set -e __getopts_optional
+    set -e __getopts_strict_mode
+    functions -e __getopts_increase_index
```
bpinto: Why do you need to remove the function?

plugins/getopts/getopts.fish
```diff
@@ -0,0 +1,399 @@
...
+        switch $token
+          case :
+            set -g __getopts_required $__getopts_required $split_string
+            set -e __getopts_required[-1]
+          case \^
+            set -g __getopts_optional $__getopts_optional $split_string
+            set -e __getopts_optional[-1]
+        end
+      end
+      functions -e __getopts_collect_optargs
```
bpinto: A function that erases itself? O.o
This is becoming too complicated.

plugins/getopts/getopts.fish
```diff
@@ -0,0 +1,399 @@
...
+              # take optional required arguments.
+              if [ -z "$suspend_break" ]
+                printf "-%s\n" $char
+              else
+                printf "%s" $char
+              end
+
+              # Suspend option break-up if the current option is either
+              # a required or an optional option type.
+              contains -- $char $__getopts_required $__getopts_optional
```
bpinto: I thought you had enjoyed using `if` statements.

plugins/getopts/getopts.spec.fish
```diff
@@ -0,0 +1,227 @@
...
+function describe_getotps -d "fish getopts"
+  function it_returns_1_if_empty
+    getopts
+    expect $status --to-equal 1
+  end
+
+  function it_reads_a_single_short_option
+    set option_string "a"
+    set option (getopts "$option_string" -a)
+    getopts
```
bpinto: I don't understand this duplicated getopts call.

Shouldn't it be something like this:

```
function it_reads_a_single_short_option
  set optarg (getopts "a:" value)
  expect $optarg[2] --to-equal a
end
```

plugins/getopts/getopts.spec.fish
```diff
@@ -0,0 +1,227 @@
...
+    expect "$option" --to-equal a
+  end
+
+  function it_reads_multiple_short_options
+    set option_string "a b c"
+    set args -ab -cb
+    set option (getopts "$option_string" $args)
```
bpinto: Isn't it the other way around?
`getopts $args "$option_string"`

bpinto:
I gave up on reviewing, can't add all the logic to my head without playing with the code. So I'm going directly to the tests.

plugins/getopts/getopts.spec.fish
```diff
@@ -0,0 +1,227 @@
...
+
+    set option (getopts "$option_string" $args)
+    getopts
+    expect "$option" --to-equal b
+  end
+
+  function it_reads_a_short_style_option
+    set option_string "a:an-option"
+    set option (getopts "$option_string" -a)
+    getopts
```
bpinto: I really don't understand this double `getopts` call, this is different from what we say on the documentation.

plugins/getopts/getopts.spec.fish
```diff
@@ -0,0 +1,227 @@
...
+
+  function it_reads_a_single_short_option
+    set option_string "a"
+    set option (getopts "$option_string" -a)
+    getopts
+    expect "$option" --to-equal a
+  end
+
+  function it_reads_multiple_short_options
+    set option_string "a b c"
```
bpinto: How can an option have no `:`?

plugins/getopts/getopts.fish
bucaran: Nope. This is how you get `utility -cb3`, etc. For example `head -c1` and `head -c 1` are both valid for options for `head` because `-c`'s argument _is_ __required__.

On the other hand, for options taking __optional__ arguments, `-<letter>value` will assign `value` as the argument to the option identified with the letter `<letter>` while `-<letter> value` (with a space in between) will parse `value` separately.

plugins/getopts/getopts.fish
bucaran: Yup. :smile: 

It's nice when you do stuff like `echo "    Hello" | cut -c5-`.

plugins/getopts/getopts.fish
bucaran: `getopts` is meant to be used inside a `while set option (getopts "optstr" argv...) ; ...; end`.

I will make sure to make this stand out more in the documentation.

plugins/getopts/getopts.fish
bucaran: Because we don't need it anymore. It's not just about releasing what you don't need, it also declares the intention that we are done with this function.

plugins/getopts/getopts.fish
bucaran: This is the closest we get to [private functions](http://stackoverflow.com/questions/25088699/make-fish-functions-private) in fish.

plugins/getopts/getopts.fish
bucaran: I do, I do. This, well, but this improves readability IMO.

plugins/getopts/getopts.spec.fish
bucaran: Because only the short-style of the option is specified.

plugins/getopts/getopts.spec.fish
bucaran: Already explained before. Nothing special here. Since the loop is granted to exit after two calls I am simply writing `getopts` twice so that I don't have to write a `while` loop for each test.

plugins/getopts/getopts.spec.fish
bucaran: It's like this `getopts <option string> [ARGV...]`

plugins/getopts/getopts.spec.fish
bucaran: This is only for the tests, because I _didn't_ want to write a `while` loop inside each test. The result is the same. In a real application you wouldn't write it like this though. In a real application you would use `while set opion (getopts <option string> [ARGV...])`.

plugins/getopts/getopts.fish
bpinto: lol

plugins/getopts/getopts.spec.fish
bpinto: Better to have tests that test what we will be using on real code?

I find tests are good to see how to use a library, and if we want people to use `while`, this won't help.

plugins/getopts/getopts.spec.fish
bpinto: I don't see this explained on the readme.

`<option string> [:][<letter>:][<word>[:[^]]]`

We dont' have a [<letter>] for instance, and I couldn't see this usage on the notes after this code.

plugins/getopts/getopts.spec.fish
bucaran: You are right about these tests won't help someone trying to learn how to use `getopts`. These are important though, because they help me test the basic functionality. I can add more tests like the last one that looks more like real code.

plugins/getopts/getopts.spec.fish
bucaran: You are right, how can this be expressed though?

`<option string> [:][<letter>][:][<word>[:[^]]]`?

plugins/getopts/getopts.fish
bpinto: I am fine with it, but it is something I'd never do.

plugins/getopts/getopts.spec.fish
bpinto: Yep, now I understand the tests, as I didn't read the code, this didn't make sense :)

plugins/getopts/getopts.fish
bucaran: fish it's a quirky little language. I wouldn't do this if this was Python, or something more sensible. Not even sure you could even do this over there.

plugins/getopts/getopts.spec.fish
bucaran: BTW I think you are right about the tests being a great way to help users know how to use a product, so I am rewriting the whole thing (the tests). I am using a helper function to wrap a call to `getopts` that exits correctly, but other than that it should be _real_ world usage.

bucaran:
@bpinto 

+ Fixed a bug in which `strict-mode` was being handled incorrectly.
+ Rewrote tests.

Let me know if there is anything else.

bpinto:
A good idea would be to update your PR description and add it as a README.md

plugins/getopts/getopts.spec.fish
```diff
@@ -0,0 +1,130 @@
...
+    set -l options $argv[1]
+    set -e argv[1]
+    while set option (getopts $options $argv)
+      if [ (count $option) -gt 1 ] # w/ argument
+        set result "$result$option[1]($option[2])"
+      else
+        set result "$result$option"
+      end
+    end
+    printf "$result$option"
```
bpinto: It would be nice to have a separator between each tuple.

plugins/getopts/getopts.spec.fish
```diff
@@ -0,0 +1,130 @@
...
+  end
+
+  function it_handles_operands
+    expect (__getopts "" 1 2 3 4 5) --to-equal "1 2 3 4 5"
+  end
+
+  function it_handles_options_and_operands
+    expect (__getopts "g:get p:put u:update d:delete t:post" \
+    -ud --put --get --post url1 url2 user:password) \
+      --to-equal "udpgturl1 url2 user:password"
```
bpinto: I thought `url1 url2 user:password` was not going to fetched as no options specify a argument (either optional or required).

bpinto: Only on strict_mode... Makes sense.

bucaran: `url1` is `-t`'s parameter. `url2` does not look like an option so `getopts` assumes we have reached the end of arguments and everything else, i.e., `url2 user:password` is fetched as _operands_.

plugins/getopts/getopts.spec.fish
```diff
@@ -0,0 +1,130 @@
...
+  function it_detects_end_of_args
+    expect (__getopts "a b c" -abc 1 2 3 -x) --to-equal "abc1 2 3 -x"
+  end
+
+  function it_returns_unknown_option_error_on_strict_mode
+    while set option (getopts ":a b c d e f" -abcXdef 1 2 3)
+      set result "$result$option"
+    end
+
+    expect $status --to-equal 3
```
bpinto: On test `it_returns_argument_missing_error_if_argument_is_missing`, you only test status, perhaps you should do the same here? And remove the expectations below?

plugins/getopts/getopts.spec.fish
bucaran: `getopts` returns the option as the first argument of the list and the options's _argument_ (if there is any) as the second argument of the list. So, the separator is `\n`

plugins/getopts/getopts.spec.fish
bucaran: I see. I still want to somehow test this part, so let me add a similar check to `it_returns_argument_missing_error_if_argument_is_missing` too and change the name a bit to reflect this.

plugins/getopts/getopts.spec.fish
bpinto: I thought that for url1 to be t parameters, as we used the long version, we had to do --post=url1

plugins/getopts/getopts.spec.fish
bpinto: Okay, comments are evil. lol

bucaran: You always end up teaching something about unit testing. You are right, this may not be the case, but if the test went and ended up failing because of this unrelated check, it would be difficult to assess where the problem originated. I will add a separate test for this.

bpinto: :+1: :tada:

plugins/getopts/getopts.spec.fish
bucaran: In this test, url1 is not a parameter of any options, they are simply operands so `getopts` sends them to stdout after successfully returning 1 to exit the while loop.

> I thought that for url1 to be t parameters, as we used the long version, we had to do --post=url1

This only applies to options in long style that take an optional argument. For long options taking a required argument you only have to `--long-option argument`.

bucaran: |        | Short     | Long     | 
|:--------:|-----------|:--------:|
| Required | -s1 -s 1  | --long 1 |
| Optional | -s1       | --long=3 |

bucaran:
Nice table added to synopsis.

bpinto:
All good?

bucaran:
:+1: 

All good. I hope future plugins start using `getopts` instead of parsing `$argv` themselves. I will definitely use it for my own-submitted plugins.

bucaran:
Thanks! :+1:

bpinto:
I will use it, but in a restricted mode. Only long mode probably. :P 
Thanks for this, great addition.

bucaran:
Of course, there is no need to create more cryptic command line utilities. This `getopts` however, is compatible with the `getopts` standard which was also my original intention.

bpinto:
@bucaran tests are broken now. =/

bucaran:
Locally? Travis?

bpinto:
Travis, master is broken.

bucaran:
I just saw it. 

I will find the culprit.

bucaran:
As you can see [here](https://github.com/bucaran/oh-my-fish/compare/f3971521...845ddfdca1), there are no changes across commits. Could this be a travis bug...?

bpinto:
I'd like to say that, but I don't think so. =/

bucaran:
@bpinto Are your tests passing?

bpinto:
The tests that are failing ar failing on Linux, I didn't run the tests on a Linux machine.

bucaran:
They were definitely passing before. I am trying to find that commit.

bpinto:
Meanwhile I'm building a dockerfile so we can run tests on linux locally.

PR #340, opened 2015-01-21
Theme Edan: Update with optional "user & host" info.

amio:
Switch using `edan-remote` or `edan-local`.

### DIFF
--- a/themes/edan/edan-local.fish
+++ b/themes/edan/edan-local.fish
@@ -1,0 +1,5 @@
+# Switch environment to "local" for edan.
+
+function edan-local
+  set -U EDAN_HOST_TYPE "local"
+end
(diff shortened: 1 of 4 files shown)


themes/edan/edan-local.fish
```diff
@@ -0,0 +1,5 @@
+# Switch environment to "local" for edan.
+
+function edan-local
+  set -U EDAN_HOST_TYPE "local"
```
amio: @bpinto As `set -e EDAN_HOST_TYPE "local"` may confuse others with comment 'Switch environment to "local" for edan', I turn to using `set -U EDAN_HOST_TYPE "local"` to be consistent.

bpinto:
Fantastic!

PR #342, opened 2015-01-23
adding a dockerfile

bpinto:
```
docker build -t oh-my-fish .
docker run -t oh-my-fish
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM ubuntu:latest
+
+# Install dependencies
+RUN apt-get -y install curl git software-properties-common
+
+# Set bootstrap script environment variables
+ENV FISH_PPA=nightly-master \
+  TRAVIS_OS_NAME=linux TRAVIS_REPO_SLUG=bpinto/oh-my-fish TRAVIS_BRANCH=master
+
+# Cache script folder
+ADD script /src/script
+WORKDIR /src
+
+# Install fish
+RUN ./script/bootstrap.sh
+
+ADD . /src
+
+CMD ["fish", "./script/run-tests.fish", "--verbose"]
+


bpinto:
@bucaran with the dockerfile, I can now run the tests on linux and reproduce the failing tests.

bucaran:
Great addition. No more chronic travis fear!

bucaran:
BTW how can I run `getopts` tests only without modifying the `CMD` in the Dockerfile?

bobthecow:
`docker run foo` overrides `CMD`.

bpinto:
`docker run -i -t oh-my-fish fish "/src/plugins/getopts/getopts.spec.fish"`

bucaran:
Thanks @bobthecow @bpinto

bucaran:
Wish I didn't have to rebuild the container every time I change the code though...

bobthecow:
it should only be rebuilding from line 17 on, right?

bucaran:
Yes, that's right. 

But changes I make to `getopts.fish` are not read when I `docker run -t oh-my-fish`.

bpinto:
You don't need to rebuild.

`docker run -i -t oh-my-fish -v ~/.oh-my-fish:/src fish "/src/plugins/getopts/getopts.spec.fish"`

bucaran:
Changes I make to getopts.fish are not read when docker run -t oh-my-fish. It is using the same original image it created when I first run docker build..., changes I make to the spec/test file do seem to go through though.

bpinto:
No changes should go through, unless you add that -v argument.

bucaran:
Any thoughts

`exec: "-v": executable file not found in $PATH
FATA[0000] Error response from daemon: Cannot start container b022e83cfa5583ef3bde6cbf578274406dfa222e76bd0221a628c73f84423800: exec: "-v": executable file not found in $PATH`

bpinto:
Sorry, `-v` needs to come before `-t`

`docker run -i  -v ~/.oh-my-fish:/src -t oh-my-fish fish "/src/plugins/getopts/getopts.spec.fish"`

bucaran:
It correctly reads my updated spec, but it does not read any changes to `getopts`.

bpinto:
```
docker run -i -t oh-my-fish cat "/src/plugins/getopts/getopts.fish"
docker run -i -t oh-my-fish cat "/src/plugins/getopts/getopts.spec.fish"
```

Should show the files when you ran `docker build`. These files should not be modified unless you run `docker build` or `docker run` with `-v`.

bucaran:
What would be the workflow to debug / test and run?

I make some changes to either `getopts.fish` or `getopts.spec.fish`, then:

+ docker build -t oh-my-fish .
+ docker run -t oh-my-fish fish "/src/plugins/getopts/getopts.spec.fish"

I thought that would be it, but it does not read changes from `getopts.fish`, only from the spec.

bpinto:
That should update `getopts.fish` too. All files inside `~/.oh-my-fish` should be updated.

Run the cat function above to confirm that the file changed after running `docker build`.

bucaran:
The cat function does show me the updated source of `getopts.fish`, but `docker run -t..` still runs the older version.

BTW, there seems to be a bug in fish `seq` when used in reverse. Anyone on linux try `seq 5 1` it does not work.

bpinto:
`seq 5 1` does not work for me either on ubuntu running on docker.

bucaran:
fish-shell/fish-shell/issues/1914

bpinto:
@bucaran these are the correct instructions now, note that in order to run with local code, you need to use `-v` option now.

Okay, the dockerfile is in a more common way now.

To build the image using ONLY data from github:
```
docker build -t oh-my-fish
```

To run the tests using ONLY data from github:
```
docker run -t oh-my-fish #Runs all tests
docker run -t oh-my-fish fish "/root/.oh-my-fish/plugins/getopts/getopts.spec.fish"
```

To run the tests using LOCAL data:
```
docker run -v (pwd):/root/.oh-my-fish -t oh-my-fish #Runs all tests
docker run -v (pwd):/root/.oh-my-fish -t oh-my-fish fish "/root/.oh-my-fish/plugins/getopts/getopts.spec.fish"
```

bucaran:
One problem is that on OS X, I don't need to specify the step increment in `seq 5 1`, but it was required on Ubuntu running on Docker. I added that and now all my tests pass locally.

The bad news is I pushed and travis is still failing.

bpinto:
Adding -1 worked on my docker machine.

bucaran:
I will create a new PR with the changes.

PR #343, opened 2015-01-23
seq call in getopts should specify negative step

bucaran:
This fixes bugs in Ubuntu running on docker. OS X tests still pass. Travis build is falling.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,4 +16,3 @@ RUN /src/script/bootstrap.sh
 WORKDIR /root/.oh-my-fish
 
 CMD ["fish", "./script/run-tests.fish", "--verbose"]
-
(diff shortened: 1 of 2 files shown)


bpinto:
Let's see if we still need to fix travis build... =/

bpinto:
We need to fix the travis workflow... =/

bucaran:
Are we running the same Linux travis is running?

bpinto:
Travis didn't fetch your changes, that is why they were failing. On master it is green now.

bucaran:
Oh, _travis_.

bpinto:
@bucaran can you add `git status` to `script/run-tests.fish` and push it to your fork? Possibly opening a pull request here? I want to confirm if the git commit is now correctly fetched.

```
#!/usr/bin/env fish

git status
set -l result 0
for test in (find $fish_path -type f -print | grep "spec.fish")
  fish $test $argv
    or set result 1
end
exit $result
```

bucaran:
@bpinto I am back. Do you still want me to do this or were you able to confirm travis is fetching what it needs to fetch?

bpinto:
I can't confirm as Travis for me has the same repo but different branch, I would like you to test it as it would be not only from a different branch but also from a different repo.

PR #363, opened 2015-02-05
add new plugin: tiny

bucaran:
# _tiny_
> git.io for the lazy

## Synopsis

```
tiny [-u --user <username>]
     [-r --repo <repository>]
     [-c --code <vanity code>]
     [-o --open]

tiny [-c --code] [-o --open] owned-repo/url
```

## Description
Tap into [git.io](http://git.io), github's URL shortener service, to easily _shorten_ any github URL. The generated URL is printed to `stdout` and copied to the clipboard. You can optionally specify the `-o` flag to open it up on your browser too.

#### `-u --user`  
Username. If omitted, the git global config is queried.

#### `-r --repo`  
Repository name.

#### `-c --code`  
Own code to setup a personal `vanity` URL. A regular short URL is generated if the specified code is not available.

#### `-o --open`  
Open the short URL in the browser.

## Default Options

It's possible to omit the `-u` option and just specify the `url` to shorten.

```fish
tiny my-awesome-repo
```

In this case, the username will be retrieved from your git global config. You can add your github username to git's config like this:

```fish
git config --global github.user <your username>
```

Another common use case is to simply copy paste an existing URL into the terminal:

```fish
tiny [-o --open] http://github.com/facebook/react
```

## Diagnostics

  The following error codes are generated:

+ _1_  the URL was invalid or git.io failed.

+ _2_  invalid input or no input.

## Links

  + [git.io](http://git.io)
  + [Github's URL shortener](https://github.com/blog/985-git-io-github-url-shortener)

## License

[MIT](http://opensource.org/licenses/MIT) © [Jorge Bucaran](https://bucaran.me)

### DIFF
--- a/plugins/tiny/README.md
+++ b/plugins/tiny/README.md
@@ -1,0 +1,73 @@
+# _tiny_
+> git.io for the lazy
+
+## Synopsis
+
+```
+tiny [-u --user <username>]
+     [-r --repo <repository>]
+     [-c --code <vanity code>]
+     [-o --open]
+     [-h --help]
+
+tiny [-c --code] [-o --open] owned-repo/url
+```
+
+## Description
+Tap into [git.io](http://git.io), github's URL shortener service, to easily _shorten_ any github URL. The generated URL is printed to `stdout` and copied to the clipboard. You can optionally specify the `-o` flag to open it up on your browser too.
+
+## Options
+
+#### `-u --user`  
+Username. If omitted, the git global config is queried.
+
+#### `-r --repo`  
+Repository name.
+
+#### `-c --code`  
+Own code to setup a personal `vanity` URL. A regular short URL is generated if the specified code is not available.
+
+#### `-o --open`  
+Open the short URL in the browser.
+
+#### `-h --help`
+Display usage help.
+
+## Default Options
+
+It's possible to omit the `-u` option and just specify the `url` to shorten.
+
+```fish
+tiny my-awesome-repo
+```
+
+In this case, the username will be retrieved from your git configuration file. You can add your github username to git's configuration like this:
+
+```fish
+git config github.user <your username>
+```
+
+Another common use case is to simply copy paste an existing URL into the terminal:
+
+```fish
+tiny [-o --open] http://github.com/facebook/react
+```
+
+## Diagnostics
+
+  The following error codes are generated:
+
++ _1_  git.io failed to shorten the URL
+
++ _2_  invalid input or no input.
+
+
+## Links
+
+  + [git.io](http://git.io)
+  + [Github's URL shortener](https://github.com/blog/985-git-io-github-url-shortener)
+
+
+## License
+
+[MIT](http://opensource.org/licenses/MIT) © [Jorge Bucaran](https://bucaran.me)
--- a/plugins/tiny/spec/tiny.spec.fish
+++ b/plugins/tiny/spec/tiny.spec.fish
@@ -1,0 +1,43 @@
+import plugins/fish-spec
+import plugins/tiny
+
+function describe_tiny -d "tiny (git.io ULR shortener)"
+  function it_generates_a_short_url
+    expect (tiny -u facebook -r react | cut -d/ -f3) --to-equal git.io
+  end
+
+  function it_prints_usage_if_there_is_no_input
+    expect (tiny | grep USAGE) --to-equal "  USAGE"
+  end
+
+  function it_returns_1_if_gitio_fails
+    tiny https://notgithub.com/bad/url
+    expect $status --to-equal 1
+  end
+
+  function it_returns_2_if_there_is_an_invalid_option
+    set -l ignore_output (tiny --wrong-option)
+    expect $status --to-equal 2
+  end
+
+  function it_returns_2_if_there_is_no_user_or_repo_specified
+    set -l ignore_output (tiny)
+    expect $status --to-equal 2
+  end
+
+  function it_prints_usage_if_there_is_an_invalid_option
+    expect (tiny -x | grep USAGE) --to-equal "  USAGE"
+  end
+
+  function it_prints_help_if_no_global_user_is_configured
+    set github_user_copy (git config --local github.user)
+
+    git config --local github.user ""
+    expect (tiny --repo whos-repo | head -c7) --to-equal "Specify"
+    expect $status --to-equal 2
+
+    git config --local github.user "$github_user_copy"
+  end
+end
+
+spec.run $argv
--- a/plugins/tiny/tiny.fish
+++ b/plugins/tiny/tiny.fish
@@ -1,0 +1,130 @@
+# NAME
+#   tiny - tap into github's git.io URL shortener
+#
+# SYNOPSIS
+#   tiny [-u --user <username>]
+#        [-r --repo <repository>]
+#        [-c --code <vanity code>]
+#        [-o --open]
+#        [-h --help]
+#
+#   tiny [-c --code] [-o --open] owned-repo/url
+#
+# DESCRIPTION
+#   Tap into git.io, github's URL shortener service, to easily shorten any
+#   github URL. The generated URL is printed to `stdout` and copied to the
+#   clipboard. You can optionally specify the `-o` flag to open it up on
+#   your browser too.
+#
+# OPTIONS
+#   -u --user  Username. If omitted, query the git global config.
+#   -r --repo  Repository.
+#   -c --code  Own code to setup a personal `vanity` URL. A regular short
+#              URL is generated if the specified code is not available.
+#   -o --open  Open the short URL in the browser.
+#   -h --help  Display usage help.
+#
+# DEFAULT OPTIONS
+#   It's possible to omit the `-u` option and just specify the `url` to
+#   shorten.
+#
+#     tiny my-awesome-repo
+#
+#   In this case, the username will be retrieved from your git config.
+#   You can add your github username to git's config like this:
+#
+#     git config github.user <your username>
+#
+#   Another common use case is to simply copy paste an existing URL into
+#   the terminal:
+#
+#     tiny [-o --open] http://github.com/facebook/react
+#
+# DIAGNOSTICS
+#   The following error codes are generated:
+#     1  git.io failed to shorten the URL
+#     2  invalid input or no input
+#
+# LINKS
+#   github.com/blog/985-git-io-github-url-shortener
+# /
+
+import plugins/getopts
+
+set -g __GITHUB_URL         "https://github.com"
+set -g __GITHUB_USER        (git config github.user)
+
+set -g __TINY_ERR_BAD_URL   1
+set -g __TINY_ERR_BAD_INPUT 2
+
+function github.io -a url code
+  if test -n "$code"
+    set code -F code=$code
+  end
+  # Extract URL from the response an remove an annoying \r at the end.
+  curl -is http://git.io/ -F url="$url" $code \
+    | grep "Location: http" | cut -c11- \
+    | awk '{print substr($0, 1, length($0) - 1)}'
+end
+
+function tiny -d "get a git.io short URL"
+  while set opts (getopts ":o:open u:user: r:repo: c:code: h:help" $argv)
+    switch $opts[1]
+      case o
+        set open true
+      case u
+        set user $opts[2]
+      case r
+        set repo $opts[2]
+      case c
+        set code $opts[2]
+      case h
+        tiny.help
+    end
+  end
+
+  test $status -gt 1
+    and tiny.help
+    and return $__TINY_ERR_BAD_INPUT
+
+  if echo "$opts" | grep --silent "http"
+    set url "$opts"
+  else if test -n "$opts"
+    # To allow the covenience of `tiny [OPTS] repo` → `tiny [OPTS] -r repo`
+    set repo "$opts"
+  end
+
+  if not set -q url
+    if not set -q user
+      if not set -q repo
+        tiny.help
+        return $__TINY_ERR_BAD_INPUT
+      end
+
+      set user $__GITHUB_USER
+      if test -z "$user"
+        echo "Specify a username via the -u option or save it to your git config."
+        echo -sn (set_color ccc) "e.g., git config --global github.user"
+        echo (set_color yellow) "username"
+        return $__TINY_ERR_BAD_INPUT
+      end
+    end
+
+    set url "$__GITHUB_URL/$user/$repo" "$code"
+  end
+
+  set url (github.io $url)
+  if test -n "$url"
+    if test (uname) != "Darwin"
+      import plugins/pbcopy
+    end
+    # Print to stdout, copy to stderr pipe it into pbcopy.
+    echo "$url" | tee /dev/stderr ^| pbcopy ^/dev/null
+
+    set -q open
+      and open "$url"
+      or true
+  else
+    return $__TINY_ERR_BAD_URL
+  end
+end
--- a/plugins/tiny/tiny.help.fish
+++ b/plugins/tiny/tiny.help.fish
@@ -1,0 +1,12 @@
+function tiny.help
+  echo "
+  USAGE
+    tiny [-u --user <username>]
+         [-r --repo <repository>]
+         [-c --code <vanity code>]
+         [-o --open]
+         [-h --help]
+
+    tiny [-c --code] [-o --open] owned-repo/url
+  "
+end


bucaran:
Thanks to @unbalancedparentheses for `pbcopy` for Linux. I am using it here to support both Mac and Linux.

plugins/tiny/README.md
```diff
@@ -0,0 +1,69 @@
...
+tiny [-o --open] http://github.com/facebook/react
+```
+
+## Diagnostics
+
+  The following error codes are generated:
+
++ _1_  the URL was invalid or git.io failed.
+
++ _2_  invalid input or no input.
```
bpinto: Test this scenario?

plugins/tiny/tiny.fish
```diff
@@ -0,0 +1,142 @@
...
+import plugins/getopts
+
+set -l __TINY_ERR_BAD_URL   1
+set -l __TINY_ERR_BAD_INPUT 2
+
+function github.get.user
+  git config --global github.user
+end
+
+function github.get.url
```
bpinto: Is this a constant? `set -l`?

plugins/tiny/tiny.fish
```diff
@@ -0,0 +1,142 @@
...
+#   github.com/blog/985-git-io-github-url-shortener
+# /
+
+import plugins/getopts
+
+set -l __TINY_ERR_BAD_URL   1
+set -l __TINY_ERR_BAD_INPUT 2
+
+function github.get.user
+  git config --global github.user
```
bpinto: What if I have a local github.user? Shouldn't it use it instead?

plugins/tiny/tiny.fish
```diff
@@ -0,0 +1,142 @@
...
+  if test -n "$code"
+    set code -F code=$code
+  end
+  # Extract URL from the response an remove an annoying \r at the end.
+  curl -is http://git.io/ -F url="$url" $code \
+    | grep "Location: http" | cut -c11- \
+    | awk '{print substr($0, 1, length($0) - 1)}'
+end
+
+function tiny.help
```
bpinto: You probably are trying to keep this function private, thus defining it on the same file. But I think this will cause fish to define this function every time `tiny` is called, and we could avoid this by using fish lazy load.

plugins/tiny/tiny.fish
```diff
@@ -0,0 +1,142 @@
...
+        set open true
+      case u
+        set user $opts[2]
+      case r
+        set repo $opts[2]
+      case c
+        set code $opts[2]
+      case \*
+        tiny.help
+        return $__TINY_ERR_BAD_INPUT
```
bpinto: `--help` is not a bad input.

plugins/tiny/spec/tiny.spec.fish
```diff
@@ -0,0 +1,34 @@
...
+    expect (tiny -u bucaran -r retab -c retab) --to-equal "http://git.io/retab"
+  end
+
+  function it_prints_usage_if_there_is_no_input
+    expect (tiny | grep USAGE) --to-equal "  USAGE"
+  end
+
+  function it_returns_1_if_there_is_bad_input
+    tiny https://notgithub.com/bad/url
+    expect $status --to-equal 1
```
bpinto: There are three scenarios where `exit 1` is returned, but only one is being tested.

plugins/tiny/tiny.fish
```diff
@@ -0,0 +1,142 @@
...
+      if not set -q repo
+        tiny.help
+        return $__TINY_ERR_BAD_INPUT
+      end
+
+      set user (github.get.user)
+      if test -z "$user"
+        echo "Specify a username via the -u option or save it to your git config."
+        echo (set_color ccc) "e.g., git config --global github.user"(set_color yellow) \
+        "username"
```
bpinto: Lack of indentation? I thought "username" was lost here.

P.S.: I wouldn't break the line here. :)

plugins/tiny/tiny.fish
bpinto: I believe this is also a constant.

bpinto:
`xsel: Can't open display: (null) : Inappropriate ioctl for device`

Is this a bug on the `pbcopy` plugin? Or on this plugin usage of pbcopy? I believe it's the former, if so, can we create a ticket and ping @unbalancedparentheses

plugins/tiny/tiny.fish
bucaran: > What if I have a local github.user? Shouldn't it use it instead?
I didn't think about this scenario? Do you have one? Should I check for both, but give priority to `--global`?

plugins/tiny/tiny.fish
bucaran: Wait. I thought this function would be created only once and stay in memory unless I specifically erase it or the shell session is terminated.

AFAIK, lazy loading in fish is just the function instantiation getting delayed _just_ before it's invoked.

plugins/tiny/tiny.fish
bucaran: :+1:

plugins/tiny/tiny.fish
bucaran: :+1:

bucaran:
The tests pass on Linux, but we are still getting `xsel: Can't open display: (null) : Inappropriate ioctl for device`.

plugins/tiny/tiny.fish
bucaran: :+1: 
$__GITHUB_URL

plugins/tiny/README.md
bucaran: :+1:

plugins/tiny/spec/tiny.spec.fish
bucaran: There could be others, but I think the only one is 

```fish
  set url (github.io $url)
  if test -n "$url"
  # ...
  else
    return $__TINY_ERR_BAD_URL # <--- here
  end
```

bucaran:
Added more tests and separated the tests that assert `exit 2` (not 1 btw).

bucaran:
Let me know if you have any other concerns.

plugins/tiny/tiny.fish
bpinto: You are right, but you don't need to evaluate this function every time you run `tiny`, only when you call `tiny --help`. We are unnecessarily pre-evaluating a function.

As a drawback, you can not overwrite this function if needed.

bpinto:
`set -l null (tiny)` could be better read if you used `set -l ignore_output (tiny)` instead.

pbcopy
```diff
@@ -0,0 +1 @@
+http://git.io/retab
```
bpinto: Not meant to add this file?

plugins/tiny/spec/tiny.spec.fish
```diff
@@ -0,0 +1,47 @@
...
+  function it_generates_a_short_url
+    expect (tiny -u facebook -r react | cut -d/ -f3) --to-equal git.io
+  end
+
+  function it_generates_a_url_with_a_vanity_code
+    expect (tiny -u bucaran -r retab -c retab) --to-equal "http://git.io/retab"
+  end
+
+  function it_prints_usage_if_there_is_no_input
+    expect (tiny | grep USAGE) --to-equal "  USAGE"
```
bpinto: I'd rather use `--to-be-true` instead as removing or adding a space to this text shouldn't break the test. But this is me being pedantic. :)

plugins/tiny/spec/tiny.spec.fish
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  function it_returns_1_if_gitio_fails
+    tiny https://notgithub.com/bad/url
+    expect $status --to-equal 1
+  end
+
+  function it_returns_2_if_there_is_an_invalid_option
+    set -l null (tiny --wrong-option)
```
bpinto: Reading the test below, we could be receiving the status code 2 because we didn't set the user or repo. We should set it here.

I think I know a better solution, but I need to work on `fish-spec` and add a new expectation, will work on it this next week.

plugins/tiny/tiny.fish
```diff
@@ -0,0 +1,143 @@
...
+#     2  invalid input or no input
+#
+# LINKS
+#   github.com/blog/985-git-io-github-url-shortener
+# /
+
+import plugins/getopts
+
+set -g __GITHUB_URL         "https://github.com"
+set -g __GITHUB_USER        (git config --global github.user)
```
bpinto: > When reading, the values are read from the system, global and repository local configuration files by default, and options --system, --global, --local and --file <filename> can be used to tell the command to read from only that location (see FILES).

> When writing, the new value is written to the repository local configuration file by default, and options --system, --global, --file <filename> can be used to tell the command to write to that location (you can say --local but that is the default).

plugins/tiny/tiny.fish
```diff
@@ -0,0 +1,143 @@
...
+         [-c --code <vanity code>]
+         [-o --open]
+         [-h --help]
+
+    tiny [-c --code] [-o --open] owned-repo/url
+  "
+end
+
+function tiny -d "get a git.io short URL"
+  while set opts (getopts ":o:open u:user: r:repo: c:code: h:help" $argv)
```
bpinto: `getopts` is awesome!

pbcopy
bucaran: That's an example of an already generated `git.io` address that won't change. Do you want me to remove this test?

plugins/tiny/spec/tiny.spec.fish
bucaran: `--to-be-true` would be nicer, but the thing is that as you will recall:

```fish
    case --to-be-true
      eval "$expected"
      test $status -eq 0
```

checks the status, and the status is only set after running `tiny`. So it was a one-line `--to-equal` or a two-line `--to-be-true`.

plugins/tiny/tiny.fish
bucaran: I can read, but I am not sure what you want to say. 

For users with more complicated configurations, you can always use `-u` though.

plugins/tiny/tiny.fish
bucaran: :+1: :+1: :+1:

plugins/tiny/spec/tiny.spec.fish
bucaran: If the user enters an invalid option `getopts` will return 2 because I specified _strict_ mode on my `option string`:

```fish
getopts ":o:open u:user: r:repo: c:code: h:help" $argv
```

After the loop breaks:
```fish
  test $status -gt 1
    and tiny.help
    and return $__TINY_ERR_BAD_INPUT
```

plugins/tiny/spec/tiny.spec.fish
bpinto: I don't understand lol
`expect (tiny | grep USAGE) --to-be-true` does not work?

pbcopy
bpinto: We don't use/need it anywhere, why keep it?

plugins/tiny/spec/tiny.spec.fish
bpinto: You have to forget about the code when writing integration tests, it's a black box. We can do changes on the code that this test will have a false positive.

But once I have the new expectation syntax, will be easier to understand.

plugins/tiny/tiny.fish
bpinto: I am saying that if you don't specify `--global`, it will automatically fetch the right one.

plugins/tiny/spec/tiny.spec.fish
bucaran: Not like that, because `--to-be-true` just checks if `$status` is truthy.

plugins/tiny/tiny.fish
bucaran: Got it. :smile:

plugins/tiny/spec/tiny.spec.fish
bpinto:
```
.oh-my-fish:master λ echo "abc" | grep "abc"
abc
.oh-my-fish:master λ echo $status
0
.oh-my-fish:master λ echo "abc" | grep "abcd"
.oh-my-fish:master λ echo $status
1
```

Status is set.

plugins/tiny/spec/tiny.spec.fish
bucaran: I see what you mean. In this case the thing is that I am only providing two error codes for more than two error scenarios. Well. One option would be to add more exit status, but I didn't want to complicate things.

bpinto: You are right, another alternative would be do add more exit status. But let's forget about this, we will talk about it in a future PR.

pbcopy
bucaran: :+1:

plugins/tiny/spec/tiny.spec.fish
bucaran: `$status` is set, but inside `expect` before we check `$status` we evaluate `$expected` which is a string at this point, the result of `tiny --help`, `"   USAGE"`...

bucaran: The thing is, we originally made `--to-be-true` et al, to only work for the following scenario:

```fish
expect true --to-be-true
```

plugins/tiny/spec/tiny.spec.fish
bucaran: :+1: 

> new expectation syntax

What do you have in mind? :)

plugins/tiny/tiny.fish
bucaran: Mmm I see. I didn't think about overwriting it. So the it belongs in its own file you say?

bpinto: Yep.

plugins/tiny/spec/tiny.spec.fish
```diff
@@ -0,0 +1,43 @@
+import plugins/fish-spec
+import plugins/tiny
+
+function describe_tiny -d "tiny (git.io ULR shortener)"
+  function it_generates_a_short_url
+    expect (tiny -u facebook -r react | cut -d/ -f3) --to-equal git.io
+  end
+
+  function it_prints_usage_if_there_is_no_input
+    expect (tiny | grep USAGE) --to-equal "  USAGE"
```
bpinto: I guess if it was `expect tiny | grep -q USAGE --to-be-true` it would work.

As right now, this is the same as `expect "   USAGE" --to-be-true` as fish will evaluate `()` before calling `expect`.

bucaran: That'd be nice, but `tiny` is a fish function, and can't use pipes. This is different to using a file for your script, like what I did with `shark`, in which case you can use pipes.

plugins/tiny/spec/tiny.spec.fish
```diff
@@ -0,0 +1,43 @@
...
+    set -l ignore_output (tiny)
+    expect $status --to-equal 2
+  end
+
+  function it_prints_usage_if_there_is_an_invalid_option
+    expect (tiny -x | grep USAGE) --to-equal "  USAGE"
+  end
+
+  function it_prints_help_if_no_global_user_is_configured
+    set github_user_copy (git config --global github.user)
```
bpinto: As we don't specify `--global` anymore, I would change this to `--local` to make this less scary. Right now, if this expectation fail, this test will have a side effect (clearing the git user configuration) as `git config --global github.user "$github_user_copy"` won't be executed.

plugins/tiny/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Default Options
+
+It's possible to omit the `-u` option and just specify the `url` to shorten.
+
+```fish
+tiny my-awesome-repo
+```
+
+In this case, the username will be retrieved from your git global config. You can add your github username to git's config like this:
```
bpinto: from your git config.
-global

plugins/tiny/spec/tiny.spec.fish
bpinto: I am not sure if I understand what you mean by can't use pipes.
`expect "tiny | grep -q USAGE" --to-be-true` works, but I'm not very happy with the syntax. I'd rather try to make `()` usage work.

plugins/tiny/spec/tiny.spec.fish
bucaran: :+1:

plugins/tiny/spec/tiny.spec.fish
bucaran: That's different. If you enclose it in strings, you are piping the result of `tiny` into `grep`, which is not what you had before without quotes.

bpinto: I still don't understand why it was not piping the result of `tiny` into `grep` before. But I'm fine with the test as we don't have better alternatives atm.

bpinto:
I think we just need to delete that `pbcopy` file and merge it. :+1:

plugins/tiny/spec/tiny.spec.fish
bucaran: In fish, you can't make a function receive input from a pipe. `isatty` will not work.

bucaran:
Going to merge.

bpinto:
@bucaran just noticed that we didn't add it to the [plugin list](https://github.com/bpinto/oh-my-fish/blob/master/plugins/README.md).

bucaran:
Sorry about that! I opened a new PR for this...

unbalancedparentheses:
@bucaran pbcopy is not working? I do not get that error message. What xsel version are you using?

bucaran:
It works, but there was some weird redirection needed to make it quiet.  I will show an example later when I get near a power station.

Mobileから送信

<div>-------- 元のメール --------</div><div>送信元： Federico Carrone <notifications@github.com> </div><div>日時:2015/02/09  AM5:37  (GMT+09:00) </div><div>宛先： bpinto/oh-my-fish <oh-my-fish@noreply.github.com> </div><div>Cc： Jorge Bucaran <jbucaran@me.com> </div><div>件名: Re: [oh-my-fish] add new plugin: tiny (#363) </div><div>
</div>@bucaran pbcopy is not working? I do not get that error message. What xsel version are you using?

bucaran:
@unbalancedparentheses 

```fish
echo "$url" | tee /dev/stderr ^| pbcopy ^/dev/null
```

This code works both on Mac and Linux, but it gives an error on Linux `xsel: Can't open display: (null) : Inappropriate ioctl for device` if I don't go out of my way to redirect the output like 

```fish
set -l ignore_output (my_utility [OPTIONS] ...)
```

This was only bugging us during tests, but the above fixed it. That however, is not as expressive as

```
my_utility [OPTIONS]... >1 ^&2
```

Issue #147, opened 2014-04-24
Option for fish's default theme


ifonefox:
Unless I'm missing something, once I change fish's theme to a custom one, I cannot change it back to the vanilla fish theme. I've tried loading a blank theme, but that doesn't bring it back either. Changing it via fish_config doesn't work too. Right now, the only way is to manually copy the theme, and create a custom theme out of it. It would be nice if there was an easier way to change the theme to any of fish's included themes.

bucaran:
If anyone feels the need to reopen this issue, please do so, I will be closing this as I see it going nowhere in the near future.

PR #364, opened 2015-02-06
Add Fishline as a theme

0rax:
As discussed with @bucaran in this [issue](https://github.com/0rax/fishline/issues/7) I created a oh-my-fish compatible version of my prompt fishline.

It's a simple segment based prompt that use some of the powerline glyphs to get a sexy look.

It could be configured using a fish global vriable called FISH_PROMPT containing the segment name you want to load.

Hope you like it.

### DIFF
--- a/themes/fishline/fish_prompt.fish
+++ b/themes/fishline/fish_prompt.fish
@@ -1,0 +1,23 @@
+# name: Fishline
+# function _update_fishline
+
+function __grab_fishline
+  set -l fpath $fish_path/themes/fishline/fishline
+  curl -sLo $fpath.zip "https://github.com/orax/fishline/archive/master.zip"
+  rmdir $fpath
+  unzip -q $fpath.zip
+  mv "fishline-master" $fish_path
+end
+
+if not test -f $fish_path/themes/fishline/fishline/fishline.fish
+  __grab_fishline
+end
+set FLINE_PATH $fish_path/themes/fishline/fishline/
+set FLINE_PROMPT STATUS JOBS PWD GIT WRITE N ROOT
+source $FLINE_PATH/themes/default.fish
+
+function fish_prompt
+
+  fishline $status
+
+end
--- a/tools/install.fish
+++ b/tools/install.fish
@@ -59,7 +59,7 @@ end
 # Either git clone or curl GET repository.
 log blue "Cloning Oh My Fish from remote repository..."
 if type git >/dev/null
-  git clone -b $TRAVIS_BRANCH "https://github.com/$TRAVIS_REPO_SLUG.git" $fish_path
+  git clone --recursive -b $TRAVIS_BRANCH "https://github.com/$TRAVIS_REPO_SLUG.git" $fish_path
 else
   log yellow "Install git to pull Oh-My-Fish updates"
   log white "Downloading remote zip from Github..."
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "themes/fishline/fishline"]
+	path = themes/fishline/fishline
+	url = https://github.com/0rax/fishline
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ We are hoping to collect a bunch of themes for our command prompts. You can see 
 
 1. Clone the repository
 ```fish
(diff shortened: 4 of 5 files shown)


tools/install.fish
```diff
@@ -59,7 +59,7 @@ end
 # Either git clone or curl GET repository.
 log blue "Cloning Oh My Fish from remote repository..."
 if type git >/dev/null
-  git clone -b $TRAVIS_BRANCH "https://github.com/$TRAVIS_REPO_SLUG.git" $fish_path
+  git clone --recursive -b $TRAVIS_BRANCH "https://github.com/$TRAVIS_REPO_SLUG.git" $fish_path
```
bucaran: This will download submodules as well correct?

Now, could there be any other side effects?

0rax: Well not really, the git manual says about the `--recursive` option:

```
       --recursive, --recurse-submodules
           After the clone is created, initialize all submodules within, using their default settings. This is
           equivalent to running git submodule update --init --recursive immediately after the clone is finished.
           This option is ignored if the cloned repository does not have a worktree/checkout (i.e. if any of
           --no-checkout/-n, --bare, or --mirror is given)
``` 
It only downloads submodule, if you make sure that no submodule are added outsides `/plugin/` and `/theme/` there is no way that unwanted code will be imported.

bucaran: Cool. This is a feature of `git` I hadn't used yet. How do I add a submodule?

0rax: It's my favorite `git` feature. It's really easy juste do
```
git submodule add 'git url'
```
It will create a `.gitmodules` in the root of the git directory.

It will be automaticaly cloned when you add it.
Just comit the folder and the `.gitmodules` file

When you clone the repository add the `--recursvie` option to clone them automaticaly or run 
```sh
git submodule init
git submodule update # with --recusrive if you want to allow submodule to automacitaly update its submodule
```

To remove a submodule just `git rm` the folder.

bpinto:
[https://github.com/bpinto/oh-my-fish/issues/259] proposes a better solution for handling external code.

I don't think we should include git submodules like that while we don't implement #259

themes/fishline/fish_prompt.fish
```diff
@@ -0,0 +1,23 @@
...
+
+function __grab_fishline
+  set -l fpath $fish_path/themes/fishline/fishline
+  curl -sLo $fpath.zip "https://github.com/orax/fishline/archive/master.zip"
+  rmdir $fpath
+  unzip -q $fpath.zip
+  mv "fishline-master" $fish_path
+end
+
+if not test -f $fish_path/themes/fishline/fishline/fishline.fish
```
bpinto: You could have concurrency issues here when opening multiple tabs without downloading this file first.

bpinto:
It's a nice theme, but we would have to do an overall refactor on the code to make it follow the common code pattern for themes. Are you interested on this challenge? :)

0rax:
Well the problem is that its not only a theme, its a framework to create your own prompt and I dont want it to be availlable only to oh-my-fish users.

I've created my own naming convetion for devs to propose themes and segments to the project. Refactoring fishline's code to be part of oh-my-fish is just out of the its concept.

That's why I proposed this solution to @bucaran in this [issue](https://github.com/0rax/fishline/issues/7).

Without a way to handle external code, I prefer not to split code databases to make a fishline like theme availlable to oh-my-fish user and have my own prompt framework as a seperate project.

bpinto:
That's fair. We do intend to implement theme components here one day, perhaps you could help with that in the future.

0rax:
I am glad to help, I will dig more into oh-my-fish internal and see what I can do.

This pull request was just a follow up to @bucaran question in the fishline repository.

I understand the need for such a framework and I will be glad to help, but as a fish dev I like to have everything seperated (I've made fishline and the port of bd to fish, and I am currently working on a autojump/jump port in pure fish).

Integrating 'third-party' code base to oh-my-fish may be a really good idea. This way devs could make their standalone fish software part of the oh-my-fish framework while having them as a separate project with their own project management.

bpinto:
:+1:

I'll close this so we can focus on the next goal. :)

References: https://github.com/bpinto/oh-my-fish/issues/147

bucaran:
@0rax Fair enough. Let me know if you have any questions.

PR #367, opened 2015-02-09
Add tab plugin.

bobthecow:
This is a port of a bash function I wrote almost five years ago:

  https://gist.github.com/bobthecow/757788

The fish version is better though ;)

I only have OS X with iTerm and Terminal to test on, but it should be possible to add support for other terminals on other OS. Send a pull request!

### DIFF
--- a/plugins/tab/tab.fish
+++ b/plugins/tab/tab.fish
@@ -1,0 +1,56 @@
+# Open new iTerm and Terminal tabs from the command line
+#
+# Author: Justin Hileman (http://justinhileman.com)
+#
+# Usage:
+#     tab                   Opens the current directory in a new tab
+#     tab [PATH]            Open PATH in a new tab
+#     tab [CMD]             Open a new tab and execute CMD
+#     tab [PATH] [CMD] ...  You can prolly guess
+
+function tab -d 'Open the current directory (or any other directory) in a new tab'
+  set -l cmd ''
+  set -l cdto $PWD
+
+  if test (count $argv) -gt 0 -a -d $argv[1]
+    pushd . >/dev/null
+    cd $argv[1]
+    set cdto $PWD
+    set -e argv[1]
+    popd >/dev/null
+  end
+
+  if test (count $argv) -gt 0
+    set cmd "; $argv"
+  end
+
+  switch $TERM_PROGRAM
+
+  case 'iTerm.app'
+    osascript 2>/dev/null -e "
+      tell application \"iTerm\"
+        tell current terminal
+          launch session \"Default Session\"
+          tell the last session
+            write text \"cd \\\"$cdto\\\"$cmd\"
+          end tell
+        end tell
+      end tell
+    "
+
+  case 'Apple_Terminal'
+    osascript 2>/dev/null -e "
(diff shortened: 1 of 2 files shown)


plugins/tab/tab.fish
```diff
@@ -0,0 +1,58 @@
...
+#     tab [PATH]            Open PATH in a new tab
+#     tab [CMD]             Open a new tab and execute CMD
+#     tab [PATH] [CMD] ...  You can prolly guess
+
+function tab -d 'Open the current directory (or any other directory) in a new tab'
+  set -l cmd ''
+  set -l cdto (pwd)
+
+  if test (count $argv) -gt 0
+    pushd . >/dev/null
```
bucaran: This pushes the current directory to the directory stack right? But why?

bobthecow: Yep :)

bobthecow: That way I let fish resolve relative paths and symlinks and all that for me, rather than trying to sort it out myself.

bobthecow: Updated this whole block in 6c85e8d to make it a bit clearer what's happening.

plugins/tab/tab.fish
```diff
@@ -0,0 +1,56 @@
...
+#     tab [CMD]             Open a new tab and execute CMD
+#     tab [PATH] [CMD] ...  You can prolly guess
+
+function tab -d 'Open the current directory (or any other directory) in a new tab'
+  set -l cmd ''
+  set -l cdto (pwd)
+
+  if test (count $argv) -gt 0 -a -d $argv[1]
+    pushd . >/dev/null
+    cd $argv[1]
```
bucaran: What if `$argv[1]` is an invalid directory?

bobthecow: that's what `-a -d $argv[1]` is supposed to check.

bucaran: Right.

bucaran:
Looks good. 

@bpinto May cavil at the lack of tests, but other than that, updating the plugins README and we are good to go.

bobthecow:
I added it to the plugins README and squashed commits :)

bucaran:
:+1:

plugins/tab/tab.fish
```diff
@@ -0,0 +1,56 @@
...
+#
+# Usage:
+#     tab                   Opens the current directory in a new tab
+#     tab [PATH]            Open PATH in a new tab
+#     tab [CMD]             Open a new tab and execute CMD
+#     tab [PATH] [CMD] ...  You can prolly guess
+
+function tab -d 'Open the current directory (or any other directory) in a new tab'
+  set -l cmd ''
+  set -l cdto (pwd)
```
bpinto: Probably better to use `$PWD` instead, so we don't start another process?

bobthecow: Works for me. Fixed in aec321b

bpinto:
Awesome, thanks for this plugin @bobthecow !

bobthecow:
No problem!

PR #369, opened 2015-02-10
Change tiny declared globals to local scope.

bucaran:
Change tiny declared globals to local scope. Better not to pollute the global space.

### DIFF
--- a/plugins/tiny/tiny.fish
+++ b/plugins/tiny/tiny.fish
@@ -51,12 +51,6 @@
 
 import plugins/getopts
 
-set -g __GITHUB_URL         "https://github.com"
-set -g __GITHUB_USER        (git config github.user)
-
-set -g __TINY_ERR_BAD_URL   1
-set -g __TINY_ERR_BAD_INPUT 2
-
 function github.io -a url code
   if test -n "$code"
     set code -F code=$code
@@ -68,6 +62,12 @@ function github.io -a url code
 end
 
 function tiny -d "get a git.io short URL"
+  set -l __github_url         "https://github.com"
+  set -l __github_user        (git config github.user)
+
+  set -l __TINY_ERR_BAD_URL   1
+  set -l __TINY_ERR_BAD_INPUT 2
+
   while set opts (getopts ":o:open u:user: r:repo: c:code: h:help" $argv)
     switch $opts[1]
       case o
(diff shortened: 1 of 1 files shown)


plugins/tiny/tiny.fish
```diff
@@ -68,6 +62,12 @@ function github.io -a url code
 end
 
 function tiny -d "get a git.io short URL"
+  set -l __github_url         "https://github.com"
+  set -l __github_user        (git config github.user)
+
+  set -l __TINY_ERR_BAD_URL   1
```
bpinto: These variables need to follow a pattern.

> Variables in fish can be exported. This means the variable will be inherited by any commands started by fish. It is convention that exported variables are in uppercase and unexported variables are in lowercase.

P.S.: It's always good to wait for a code review before merging.

bobthecow: Also double underscore seems weird for local variables.

bucaran: I am not sure I _completely_ agree with that code convention. At least for error codes, it feels more natural to make them uppercase.

<div>-------- 元のメール --------</div><div>送信元： Bruno <notifications@github.com> </div><div>日時:2015/02/11  AM1:54  (GMT+09:00) </div><div>宛先： bpinto/oh-my-fish <oh-my-fish@noreply.github.com> </div><div>Cc： Jorge Bucaran <jbucaran@me.com> </div><div>件名: Re: [oh-my-fish] Change tiny declared globals to local scope. (#369) </div><div>
</div>In plugins/tiny/tiny.fish:

> @@ -68,6 +62,12 @@ function github.io -a url code
>  end
>  
>  function tiny -d "get a git.io short URL"
> +  set -l __github_url         "https://github.com"
> +  set -l __github_user        (git config github.user)
> +
> +  set -l __TINY_ERR_BAD_URL   1
These variables need to follow a pattern.

Variables in fish can be exported. This means the variable will be inherited by any commands started by fish. It is convention that exported variables are in uppercase and unexported variables are in lowercase.

P.S.: It's always good to wait for a code review before merging.

bucaran: @bobthecow I agree double underscore for _local_ variables is weird. My bad. :sweat_smile:

@bpinto I am not sure I _completely_ agree with that code convention. At least for error codes, it feels more natural to make them uppercase.

bpinto: I don't see any problems with having error codes downcased, I honestly don't enforce this pattern on plugins and themes.

If you want to follow your own convention, it's your call. But, personally, I don't think fighting with a language style guide is the way to go.

bucaran: Me neither, looking at it now I have an urge to make them lowercase. 

Having a style guide is a cool move. I don't know of any other fish style guides out there, so we could _pioneer_ there.

PR #373, opened 2015-02-11
Fix getopts multiline

bucaran:
getopts should handle multiline option strings.

See [getopts#6](https://github.com/bucaran/getopts/issues/6)


bucaran:
Change is just `tr '[:space:]' \n` to `tr -s '[:space:]' \n`. 

```
-s      Squeeze multiple occurrences of the characters listed in the last operand
             (either string1 or string2) in the input into a single instance of the char-
             acter.  This occurs after all deletion and translation is completed.
```

Nothing special.

PR #378, opened 2015-02-14
Add root .editorconfig file.

amio:
[EditorConfig](http://EditorConfig.org) should do some help to restrict coding style :)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Two-space indent, Unix-style newlines, and a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2


bobthecow:
I'd just make everything two-space indent, tbh. We can put in exceptions if we ever need 'em :)

bucaran:
@bobthecow Agree.

@amio 
Isn't there a way to combine
```
[*.fish]
indent_style = space
indent_size = 2


# 2-space for .load file
[*.load]
indent_style = space
indent_size = 2
```
?

.editorconfig
```diff
@@ -0,0 +1,19 @@
...
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+# 2-space for .fish file
+[*.fish]
```
bpinto: Does `[*.fish, *.load]` works?

bobthecow:
```ini
# EditorConfig is awesome: http://EditorConfig.org

# top-most EditorConfig file
root = true

# Two-space indent, Unix-style newlines, and a newline ending every file
[*]
end_of_line = lf
insert_final_newline = true
indent_style = space
indent_size = 2
```

.editorconfig
amio: `[ *.{fish,load} ]` will work

bucaran: :+1: Nice

amio:
@bobthecow I thought declare them separately could make rules clearly, and since I didn't know if we all prefer to use 2-space indent in most cases. Well, I know now :)

Force updated~

bpinto:
Thanks @amio! This is great.

PR #381, opened 2015-02-17
Add `update_oh_my_fish.fish`

amio:
Enable user to update OMF with 'update_oh_my_fish', by git-pull or latest zip.

### DIFF
--- a/functions/update_oh_my_fish.fish
+++ b/functions/update_oh_my_fish.fish
@@ -30,6 +30,9 @@ cd $fish_path
 
 if [ -d .git ]
 
+	# debug log
+	log yellow "To update through git."
+
   if [ ( _is_git_touched ) ]
 
     echo "Error: Your local changes would be overwritten by update.
@@ -52,16 +55,29 @@ if [ -d .git ]
 
 else
 
+	# debug log
+	log yellow "To update throught zip."
+
   find . -maxdepth 1 -not -name 'custom' | xargs rm -rf ^/dev/null
 
   log blue "Downloading remote zip from Github..."
 
   if curl -Lo $omf_temp_zip "https://github.com/$TRAVIS_REPO_SLUG/archive/master.zip"
 
-    unzip -uo $omf_temp_zip -d $omf_temp_dir >/dev/null
+    #debug log
+		log yellow "To unzip files to $omf_temp_dir"
+		unzip -uo $omf_temp_zip -d $omf_temp_dir >/dev/null
+
+
+		#debug log
+		log yellow "To copy files to $fish_path"
     cd $omf_temp_dir/oh-my-fish-master
     cp -R * $fish_path
       and log green "Oh-My-Fish succesfully downloaded and extracted to $fish_path"
+
+
+    #debug log
+		log yellow "To clean files in $omf_temp_dir"
     rm -rf $omf_temp_zip $omf_temp_dir
 
   else
--- a/functions/update_oh_my_fish.fish
+++ b/functions/update_oh_my_fish.fish
@@ -1,0 +1,75 @@
+#!/usr/bin/env fish
+#
+# Update oh-my-fish through git-pull or latest zip.
+#/
+
+# Check git repo status.
+function _is_git_touched
+	echo -n (command git status --porcelain ^/dev/null | grep "^[ MADRC]")
+end
+
+# Log a color message.
+#   log <color> <string>...
+function log
+  echo -e (set_color $argv[1])$argv[2..-1](set_color normal)
+end
+
+# Allow installers to specify the source repository.
+if not set -q TRAVIS_REPO_SLUG
+  set TRAVIS_REPO_SLUG bpinto/oh-my-fish
+  set TRAVIS_BRANCH master
+end
+
+set -l msg_done "Oh-My-Fish update successful!"
+set -l omf_temp_zip "/tmp/oh-my-fish-latest.zip"
+set -l omf_temp_dir "/tmp/omf-latest"
+
+cd $fish_path
+
+# Either git pull or curl GET repository.
+
+if [ -d .git ]
+
+  if [ ( _is_git_touched ) ]
+
+    echo "Error: Your local changes would be overwritten by update.
+       Please commit your changes or stash them before updating.
+       Update aborted."
+    exit 1
+
+  else
+
+    log blue "Pull Oh-My-Fish from remote repository..."
+
+    git checkout $TRAVIS_BRANCH
+    git pull
+    and echo $msg_done.
+
+    or log red "Pull from remote repository failed."
+    and exit 1
+
+  end
+
+else
+
+  find . -maxdepth 1 -not -name 'custom' | xargs rm -rf ^/dev/null
+
+  log blue "Downloading remote zip from Github..."
+
+  if curl -Lo $omf_temp_zip "https://github.com/$TRAVIS_REPO_SLUG/archive/master.zip"
+
+    unzip -uo $omf_temp_zip -d $omf_temp_dir >/dev/null
+    cd $omf_temp_dir/oh-my-fish-master
+    cp -R * $fish_path
+      and log green "Oh-My-Fish succesfully downloaded and extracted to $fish_path"
+    rm -rf $omf_temp_zip $omf_temp_dir
+
+  else
+
+    log red "Oh-My-Fish could not be downloaded."
+    log white "Report an issue → github.com/bpinto/oh-my-fish/issues"
+    exit 1
+
+  end
+
+end


functions/update_oh_my_fish.fish
```diff
@@ -0,0 +1,75 @@
...
+       Please commit your changes or stash them before updating.
+       Update aborted."
+    exit 1
+
+  else
+
+    log blue "Pull Oh-My-Fish from remote repository..."
+
+    git checkout $TRAVIS_BRANCH
+    git pull
```
bucaran: What if I have changes committed locally? I think we want to `pull --rebase` here right?

functions/update_oh_my_fish.fish
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env fish
+#
+# Update oh-my-fish through git-pull or latest zip.
+#/
+
+# Check git repo status.
+function _is_git_touched
+	echo -n (command git status --porcelain ^/dev/null | grep "^[ MADRC]")
```
bucaran: `command` will make sure to run `git` as opposed to a user defined function `git`, which is not all that unlikely, but what if the user is using a `git` wrapper with some extra functionality. 

Can you state what is was your motivation of using `command` here?

functions/update_oh_my_fish.fish
```diff
@@ -0,0 +1,75 @@
...
+end
+
+# Log a color message.
+#   log <color> <string>...
```
bucaran: Some of these functions are already defined in `oh-my-fish.fish`, maybe we should find a way to reuse some of them?

functions/update_oh_my_fish.fish
```diff
@@ -0,0 +1,75 @@
...
+set -l omf_temp_zip "/tmp/oh-my-fish-latest.zip"
+set -l omf_temp_dir "/tmp/omf-latest"
+
+cd $fish_path
+
+# Either git pull or curl GET repository.
+
+if [ -d .git ]
+
+  if [ ( _is_git_touched ) ]
```
bucaran: Does this work as expected or am I missing something?

`_is_git_touched` will be evaluated and expanded causing `$status` to change if it fails, but you are wrapping the actual `git status` call in an `echo` statement which would reset `$status` to 0.

functions/update_oh_my_fish.fish
```diff
@@ -0,0 +1,75 @@
...
+    and echo $msg_done.
+
+    or log red "Pull from remote repository failed."
+    and exit 1
+
+  end
+
+else
+
+  find . -maxdepth 1 -not -name 'custom' | xargs rm -rf ^/dev/null
```
bucaran: So, delete everything _except_ the `custom/` directory and overwrite? 

What about first checking if there are any new commits up on github, if so, download remote repo.

> You know what would be kind of cool? Using `osascript` to launch a new terminal tab to do the downloading in the back. I don't know :grinning: 

After that, we could get the [checksum of the directory tree](http://unix.stackexchange.com/questions/35832/how-do-i-get-the-md5-sum-of-a-directorys-contents-as-one-sum) and overwrite only what's dirty locally after a diff.

Oh, what fun!

bucaran:
Good job @amio! Please see some of the issues I raised in the comments above.

functions/update_oh_my_fish.fish
amio: I have barely used `pull --rebase`, looks it's time to make some practice now :P

functions/update_oh_my_fish.fish
amio: Ooops, I misunderstood `command` in fish docs, I thoughts it force to use native `git` command
> <

functions/update_oh_my_fish.fish
amio: It works as expected. I'm not very familiar with exact grammar of `if` but according my tests, it seems accept the standard output rather than exit code.

functions/update_oh_my_fish.fish
amio: Sure~

functions/update_oh_my_fish.fish
amio: Question 1, yes, preserve `custom` directory during updating.

For `osascript`, I think oh-my-fish may run on linux too, so we should find a way more compatible?
To diff between two local directory seems a little heavy for an updating? Anyway, a simple copy works well too :P

bucaran: Indeed.

I am not sure about that. Would not a diff be rather light?

functions/update_oh_my_fish.fish
bucaran: :)

functions/update_oh_my_fish.fish
bpinto: I think you are right on using `command git` as you want to use the native `git` command.

functions/update_oh_my_fish.fish
```diff
@@ -0,0 +1,75 @@
...
+end
+
+# Log a color message.
+#   log <color> <string>...
+function log
+  echo -e (set_color $argv[1])$argv[2..-1](set_color normal)
+end
+
+# Allow installers to specify the source repository.
+if not set -q TRAVIS_REPO_SLUG
```
bpinto: Why do you need this? If it is for supporting update from a remote zip that is not this repo, I don't think we should support it.

functions/update_oh_my_fish.fish
```diff
@@ -0,0 +1,75 @@
...
+
+cd $fish_path
+
+# Either git pull or curl GET repository.
+
+if [ -d .git ]
+
+  if [ ( _is_git_touched ) ]
+
+    echo "Error: Your local changes would be overwritten by update.
```
bpinto: We should send this to stderr.

functions/update_oh_my_fish.fish
```diff
@@ -0,0 +1,75 @@
...
+       Please commit your changes or stash them before updating.
+       Update aborted."
+    exit 1
+
+  else
+
+    log blue "Pull Oh-My-Fish from remote repository..."
+
+    git checkout $TRAVIS_BRANCH
```
bpinto: We don't need this parameter. `git pull --rebase` should be enough.

functions/update_oh_my_fish.fish
```diff
@@ -0,0 +1,75 @@
...
+    git pull
+    and echo $msg_done.
+
+    or log red "Pull from remote repository failed."
```
bpinto: These `and/or` flow are too hard to understand, please use `if` statements.

functions/update_oh_my_fish.fish
bpinto: We probably need to copy these to another folder so that it's not overwritten by the download zip file.
e.g.: One could change the `example.load` file.

functions/update_oh_my_fish.fish
```diff
@@ -0,0 +1,75 @@
...
+
+    unzip -uo $omf_temp_zip -d $omf_temp_dir >/dev/null
+    cd $omf_temp_dir/oh-my-fish-master
+    cp -R * $fish_path
+      and log green "Oh-My-Fish succesfully downloaded and extracted to $fish_path"
+    rm -rf $omf_temp_zip $omf_temp_dir
+
+  else
+
+    log red "Oh-My-Fish could not be downloaded."
```
bpinto: These should output to stderr too.

bucaran:
@amio @bpinto Just a thought. Wouldn't this be better in the form a plugin `update`?

bpinto:
I believe this should be part of the framework, as installing is part of it.

functions/update_oh_my_fish.fish
amio: Aha, indeed. I just was to follow the coding style. Sorry for copy & paste without noticed the comment :P

functions/update_oh_my_fish.fish
amio: OK~

functions/update_oh_my_fish.fish
amio: Maybe user just forgot to check back to master? For me, personally I prefer to keep the official code in master, and for most people, I guess, If someone have committed some modification to an other branch rather than master, he probably means to contribute to oh-my-fish, not just merge his custom things with official branch.

functions/update_oh_my_fish.fish
amio: I have same feeling too ;)

amio:
A plugin needs to be enabled for using, well update & installing should be available by default I think,
I agree with @bpinto

functions/update_oh_my_fish.fish
amio: Sure.

functions/update_oh_my_fish.fish
bpinto: Then perhaps we need to check if he is on master branch and halt the update if he is not?

functions/update_oh_my_fish.fish
bpinto: :+1: :)

functions/update_oh_my_fish.fish
amio: Well, the `log` function only appears in `install.fish`, should I source that in `update_oh_my_fish.fish` or move it to a standalone script?

bucaran:
Hi @amio! Do you think it would be possible to add tests to this. You don't have to use `fish-spec`, but a debug version (that just echoes each step of the process) would be helpful to speed up the merging.

amio:
No problem~

bucaran:
Wow, that was quick. Thanks and let me know if you have any questions.

amio:
Sorry for the delay, really busy these days >.<

I've added some debug log, If anything is still unclear please let me know :D

functions/update_oh_my_fish.fish
bucaran: What happened to this part?

bpinto:
Are you going to go through the comments @amio ?

amio:
I hope to but may not very quickly, it will take some time for me to learn more shell scripting but I'm little buzy this month, so if @bucaran can (he certainly can ^^ ) finish it soon, just go ahead~ And if it's no hurry, I'm surely glad to do it myself :)

bucaran:
@amio :+1:

bucaran:
I've been thinking about this, and although it was me who originally introduced the idea of downloading the ZIP and giving support to non-git users, now it really doesn't make sense to support non-git users. Really, if you want to try oh-my-fish, but don't have `git` installed you are doing something wrong.

amio:
That make a few things simpler a lot :D 
Then the updating script maybe better starting fresh.

bpinto:
@amio if you want to add more features to `omf self-update`, that would be awesome!

amio:
Cool :+1: 
I love the 'git-everywhere' idea :D

PR #384, opened 2015-02-19
Fix import for completions

bpinto:


bpinto:
@bucaran see how these tests fail

bpinto:
Can we solve this without changing `_prepend_tree` to ignore completions folder?

bucaran:
If this tests should pass, merge, and then we can fix import.

bpinto:
Tests do not pass.

Expectation:
it_does_not_import_completion_files
it_adds_completion_files_to_completion_path

bpinto:
Currently we import files under completion folder by adding them to fish_function_path.

bucaran:
I see, so we should not import files inside `completions`?

So, how do completions work then? Does fish check any directories named `completions` and loads them automatically or something?

bpinto:
AFAIK, it is similar to how function works, if you want the completions for a command called theme, it will look for a `theme.fish` file on `fish_complete_path` env var.

bpinto:
Do you have a better solution to this problem then the one on the last commit?

functions/_prepend_tree.fish
```diff
@@ -84,13 +84,9 @@ function _prepend_tree -d "Add a dependency tree to the Fish path."
...
 
-  # Null wildcard expansion will break the for loop even if $path is valid.
-  # $subs will become an empty list for directories without sub directories
-  # which is safe to use in the loop.
-  set -l subs $path/**/
-
-  # Traverse $path and $subs prepending only directories with matches.
-  for dir in $path $subs
+  # Traverse $path prepending only directories with matches. Excludes completions folder.
+  test -d $path
```
bucaran: Why do we need to exclude completion folders?

bucaran: Nevermind. All is clear now.

bucaran:
Another feature of `fish` I didn't understand clearly. This is an awesome feature by the way.

bpinto:
Fish is full of hidden gems. :+1:

bucaran:
The solutions does seem a bit hacky, maybe we can come up with something cleaner? Let me see.

bpinto:
You only say that now? After I merged? :(

I hope you find something cleaner.

bucaran:
Sorry, it took my brain thousands of nanoseconds to finally grasp fish completions.

bucaran:
To recapitulate:

> AFAIK, it is similar to how function works, if you want the completions for a command called theme, it will look for a theme.fish file on fish_complete_path env var.

:+1:

bpinto:
:+1:

PR #389, opened 2015-02-22
Improvements upon pyenv plugin.

bucaran:
+ No need to modify $PATH if pyenv is already available.
  e.g., custom installations symlinking into `/usr/local`.

+ Run pyenv init to setup pyenv env, load completions, etc.
+ Reduce some code redundancy.

### DIFF
--- a/plugins/pyenv/pyenv.load
+++ b/plugins/pyenv/pyenv.load
@@ -1,7 +1,7 @@
-if test -n "$PYENV_ROOT"
-  _prepend_path $PYENV_ROOT/bin
-  _prepend_path $PYENV_ROOT/shims
-else
-  _prepend_path $HOME/.pyenv/bin
-  _prepend_path $HOME/.pyenv/shims
+if not test -q PYENV_ROOT
+  set PYENV_ROOT $HOME/.pyenv
+end
+
+if status --is-interactive
+  . (pyenv init - | psub)
 end


plugins/pyenv/pyenv.load
```diff
@@ -1,7 +1,19 @@
...
-  _prepend_path $PYENV_ROOT/bin
-  _prepend_path $PYENV_ROOT/shims
+# Set pyenv path to point to $HOME by default
+if not test -q PYENV_ROOT
+  set PYENV_ROOT $HOME/.pyenv
+end
+
+# No need to modify $PATH if pyenv is already available.
+# e.g., custom installations symlinking into /usr/local
+if type pyenv >/dev/null
```
bpinto: What if `pyenv` is only available inside `$PYENV_ROOT/bin`? Shouldn't we still call `. (pyenv init - | psub)`?

bucaran: How is `pyenv` be only available inside `$PYENV_ROOT/bin`? I don't understand that scenario.

OK, I need to check what happens if you invoke `. (pyenv init - | psub)` when `$PYENV_ROOT` points to an arbitrary location. It should be fine though, but just to make sure.

bucaran: I have been going through this again and I think this plugin makes no sense, or I am missing something. Can anybody explain to me what's wrong with just:


```fish
if not test -q PYENV_ROOT
  set PYENV_ROOT $HOME/.pyenv
end

if status --is-interactive
  . (pyenv init - | psub)
end
```

It's not necessary to add anything to the path, `pyenv init -` already handles all that, as well as:

+ Installing the shims that capture `python` calls, so that we run the correct version.
+ Add itself as well as the __currently__ selected `python`'s version downloaded packages to the `$PATH`.
+ Completions.

bucaran:
It's not necessary to add anything to the path, `pyenv init -` already handles all that, as well as:

```fish
if not test -q PYENV_ROOT
  set PYENV_ROOT $HOME/.pyenv
end

if status --is-interactive
  . (pyenv init - | psub)
end
```
+ Installs shims that capture `python` calls, so that we run the correct version.
+ Adds itself as well as the __currently__ selected `python`'s version downloaded packages to the `$PATH`.
+ Adds Completions.

PR #390, opened 2015-02-22
Fix python version detection in theme bobthefish

ccwang002:
Relax the python version matching, so `python` that symlinks to either major versions (e.g. 2, 3) or minor versions (e.g. 2.7, 3.4) can be properly detected.

On OSX it is easy to reproduce,
```fish
$ vf tmp -p python3
(tempenv) $ readlink (which python)
# python3.4
```
Same case for python 2.7.

Note that currently both pypy and pypy3 have same display glyph. I simply relaxed their case pattern to be `pypy*`, but we could consider explicitly distinguish them in future (maybe having `p2` and `p3` superscript?)

### DIFF
--- a/themes/bobthefish/fish_prompt.fish
+++ b/themes/bobthefish/fish_prompt.fish
@@ -327,11 +327,11 @@ end
 function __bobthefish_virtualenv_python_version -d 'Get current python version'
   set -l python_version (readlink (which python))
   switch "$python_version"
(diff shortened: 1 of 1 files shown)


bucaran:
Sorry about the failing tests. 

Looks good to me.

@bobthecow?

bobthecow:
:+1:

PR #391, opened 2015-02-23
Allow adding bundler plugin execs in local config.

bobthecow:
```fish
set bundler_plugin_execs jekyll mustache
set fish_plugins bundler
```

### DIFF
--- a/plugins/bundler/bundler.load
+++ b/plugins/bundler/bundler.load
@@ -1,6 +1,10 @@
 # These methods override the default calls to append `bundle exec` if this gem
 # is available within the context of bundler.
 #
+# You can add more bundled executables in your config.fish:
+#
+#     set bundler_plugin_execs jekyll mustache
+#
 set -l execs annotate      \
              cap           \
              capify        \
@@ -27,6 +31,10 @@ set -l execs annotate      \
              thor          \
              unicorn       \
              unicorn_rails
+
+if set -q bundler_plugin_execs
+  set execs $execs $bundler_plugin_execs
+end
 
 # Fish 2.1.1+ has support for --inherit-variable
 set -l do_eval (echo $FISH_VERSION | grep 2.1.1-)


bpinto:
Add more execs instead? These are not user specific executables, so the more execs we support the best for everyone.

bobthecow:
Jekyll, yeah, but Mustache isn't bundler/ruby-specific... You can install a mustache executable via npm, and any number of other package managers. Even if there are a lot that should be included, there still should be a mechanism for explicitly setting ones that aren't universally applicable.

bpinto:
If mustache is not bundler/ruby-specific why run it with `bundle exec`?

Do we still need to run commands with `bundle exec` now that binstub is the preferred way to run these commands?

bobthecow:
It's not *always* bundler/ruby-specific. In my case, there's an executable called `mustache` that I installed via rubygems. But there's also an executable in the mustache.js npm package. Hijacking that would be rude :)

plugins/bundler/bundler.load
```diff
@@ -1,6 +1,10 @@
 # These methods override the default calls to append `bundle exec` if this gem
 # is available within the context of bundler.
 #
+# You can add more bundled executables in your config.fish:
+#
+#     set bundler_plugin_execs jekyll mustache
```
bpinto: Does it need to be `set -g`?

bpinto:
:P Okay.

plugins/bundler/bundler.load
bobthecow: Doesn't seem to? I usually write that, but I realized when adding the instructions for this than nothing in my config.fish is `set -g`.

bucaran:
Errors are fixed now. Rebase please :)

bobthecow:
No problem. Rebased.

bucaran:
:+1:

PR #393, opened 2015-02-25
+ pyenv plugin should source (pyenv init - | psub).

bucaran:
```fish
if not test -q PYENV_ROOT
  set PYENV_ROOT $HOME/.pyenv
end

if status --is-interactive
  . (pyenv init - | psub)
end
```

It's not necessary to add anything to the path, `pyenv init -` already handles all that, and also:

+ Installs shims that capture `python` calls, so that we run the correct version.
+ Adds itself as well as the __currently__ selected `python`'s version downloaded packages to the `$PATH`.
+ Adds Completions.


bucaran:
@bpinto What do you think?

bucaran:
@bpinto Thoughts?

PR #394, opened 2015-02-28
Fix for better-alias plugin

daveyarwood:
`better-alias` (`balias`) wasn't working for me out-of-the-box, on OSX Mavericks 10.9.5. After some testing, it turned out that the culprit was [the version of `expr` that comes with OSX](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/expr.1.html), which does not support the `expr match str1 str2` syntax. Instead, you have to use the (apparently older) syntax, `expr str1 : str2`. (See [this question on StackOverflow](http://stackoverflow.com/questions/25587446/syntax-error-expr) for more info.) After making this change, `balias` works as expected.

I tried the `expr str1 : str2` syntax in a Linux VM, and it works, so this change should make `balias` work for Mac users without breaking things for Linux users.

### DIFF
--- a/plugins/better-alias/balias.fish
+++ b/plugins/better-alias/balias.fish
@@ -1,11 +1,10 @@
 function balias --argument alias command
   eval 'alias $alias $command'
   if expr $command : '^sudo '>/dev/null
-    set command (expr substr + $command 6 (expr length $command))
+    set command (echo "$command" | cut -c6-)
   end
   complete -c $alias -xa "(
     set -l cmd (commandline -pc | sed -e 's/^ *\S\+ *//' );
     complete -C\"$command \$cmd\";
   )"
 end
-
--- a/plugins/better-alias/balias.fish
+++ b/plugins/better-alias/balias.fish
@@ -1,6 +1,6 @@
 function balias --argument alias command
   eval 'alias $alias $command'
-  if expr match $command '^sudo '>/dev/null
+  if expr $command : '^sudo '>/dev/null
     set command (expr substr + $command 6 (expr length $command))
   end
   complete -c $alias -xa "(
--- a/plugins/better-alias/balias.spec.fish
+++ b/plugins/better-alias/balias.spec.fish
@@ -5,6 +5,7 @@ function describe_library -d "better-alias"
 
   function after_all
     functions -e changedir
+    functions -e ls_as_root
   end
 
   function it_doesnt_fail
@@ -23,6 +24,11 @@ function describe_library -d "better-alias"
     expect test $status --to-be-true
     cd ..
     rmdir testdir
+  end
+
+  function it_chops_off_sudo
+    balias ls_as_root 'sudo ls'
+    expect test $status --to-be-true
   end
 
 end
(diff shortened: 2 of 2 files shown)


bucaran:
It would be useful to have simple [tests](https://github.com/bpinto/oh-my-fish/blob/master/plugins/fish-spec/README.md) to go along with this. BTW why do you use `balias` for?

daveyarwood:
Good call -- I'll see what I can do.

I have this as part of my `config.fish`:

```
##### git shortcuts #####

balias git hub
balias g git
balias gs 'git status'
balias gc 'git commit'
balias gco 'git checkout'
function ga; git add $argv; git status; end
balias gl "git log --graph --pretty=format:\"%C(yellow)%h %ad%Cred%d %Creset%s%Cblue [%cn]\" --decorate --all --date=short"
balias gd 'git diff'
balias gp 'git push'
balias gpr 'git pull-request'
```
Just little shortcuts for my most frequent git activities. `balias` is nice because it preserves autocompletions, so for example, I can type `gco ` and then hit tab and see all of the branches I can check out.

daveyarwood:
I took a stab at writing some simple tests using fish-spec. I couldn't figure out a good way to test that the autocompletions work, unfortunately, but something is still better than nothing, right? :smiley_cat:

bucaran:
Thanks.

You correctly fixed the previous syntax mess, so that:

```fish
if expr "sudo " : '^sudo '>/dev/null
  echo ok
end
```

Passes. But if the user tried to create an `alias` with `sudo ...` then it will fail because:
```fish
set command (expr substr + $command 6 (expr length $command))
```

Check this out:

```
 According to the POSIX standard, the use of string arguments length, substr, index, or match produces undefined results. In this version of expr, these arguments are treated just as their respective string values.
```

bucaran:
For posterity, in the `fish` [`git` completion file](https://github.com/fish-shell/fish-shell/blob/master/share/completions/git.fish), there is a function that will check git aliases. This is used for providing completions to aliases as well. 

```fish
function __fish_git_aliases
    set -l IFS \n
    command git config -z --get-regexp '^alias\.' | while read -lz key value
        begin
            set -l IFS "."
            echo -n $key | read -l _ name
            echo $name
        end
    end
end
```

This will work for `<command>=<alias>` git aliases in gitconfig, but not for an alias of the `git` command.

See fish-shell/fish-shell#1962

bucaran:
> It worked for me when I used the command test, like so:
@daveyarwood Thanks for reporting this!

bucaran:
@daveyarwood I guess you are not using `balias` with `sudo` statements, because `set command (expr substr + $command 6 (expr length $command))` does not work. 

What do you think we either remove that or add something like `set command (echo "$command" | cut -c6-)`

daveyarwood:
`balias` seems like a good workaround for now for the git aliases issue.

Good catch on the `sudo` thing -- it's probably another bash version issue between Linux and OSX. `echo "$command" | cut -c6-)` is a good fix -- it works for me both in OSX and my Linux VM.

bucaran:
:+1:

PR #398, opened 2015-03-05
Add emacs plugin

oneKelvinSmith:
Port of oh-my-zsh emacs plugin.
Includes a function to kill emacs from command line once
a daemon has started.

More information included in plugins/emacs/README.md.

See reference plugin here:
https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/emacs

### DIFF
--- a/plugins/emacs/functions/ecd.fish
+++ b/plugins/emacs/functions/ecd.fish
@@ -1,0 +1,9 @@
+function ecd
+  set -l cmd '(let ((buf-name (buffer-file-name (window-buffer))))
+             (if buf-name (file-name-directory buf-name)))'
+  set -l dir (__launch_emacs --eval $cmd | tr -d '\"')
+
+  test -n $dir
+  and echo $dir
+  or echo 'cannot deduce current buffer filename.' >/dev/stderr; return 1;
+end
--- a/plugins/emacs/spec/emacs.spec.fish
+++ b/plugins/emacs/spec/emacs.spec.fish
@@ -1,0 +1,88 @@
+set spec_path (dirname (status --current-filename))
+set plugins_path (cd $spec_path; and cd ../..; and pwd)
+set load_file $plugins_path/emacs/emacs.load
+
+import plugins/fish-spec
+
+function describe_emacs
+
+  function it_sets_editor_on_load    
+    set -e EDITOR
+    set emacs '/bin/emacs'
+    set emacs_version 25
+
+    source $load_file
+
+    expect $EDITOR --to-equal 'emacs'
+  end
+
+  function it_does_not_set_editor_when_it_is_already_set
+    set emacs '/bin/emacs'
+    set emacs_version 25
+    
+    set EDITOR 'vim'
+    
+    source $load_file
+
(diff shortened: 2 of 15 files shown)


bucaran:
Awesome. I am a `vim` person, but I use `emacs` bindings so there you go.

Did you know we have `before_all` and `after_all` in the spec? I think you don't have to declare any functions at the top of your spec. I will go over this later when I have more time.

Thanks for the contribution :+1:

oneKelvinSmith:
So, I have found a few issues with what I've submitted. Namely that some of the functions are missing the `$argv` which means that emacs won't read in filenames given at the command line. 

I also hadn't noticed that the plugin changes directory to `$fish_path/plugins/emacs` every time a new session is started.

I need to do some more work on this and submit another request a bit later. I'll take a look at `before_all` and `after_all` as well.

Sorry about this.

bucaran:
My fault as well since I didn't code review together. Let me take a look at it now and leave some comments.

plugins/emacs/functions/ecd.fish
```diff
@@ -0,0 +1,9 @@
+function ecd
+  set -l cmd '(let ((buf-name (buffer-file-name (window-buffer))))
+             (if buf-name (file-name-directory buf-name)))'
+  set -l dir (__launch_emacs --eval $cmd | tr -d '\"')
+
+  test -n $dir
```
bucaran: Let's break this up using more readable `if` conditionals. I used to like hacking with `and` and `or`, but this is hard to read in the long run.

```
if test -n $dir
  echo $dir
else
  ...
end
```

plugins/emacs/functions/ecd.fish
```diff
@@ -0,0 +1,9 @@
...
+  and echo $dir
+  or echo 'cannot deduce current buffer filename.' >/dev/stderr; return 1;
```
bucaran: We don't charge per line usage. Let's use one line per statement. :grinning:

plugins/emacs/spec/emacs.spec.fish
```diff
@@ -0,0 +1,88 @@
+set spec_path (dirname (status --current-filename))
+set plugins_path (cd $spec_path; and cd ../..; and pwd)
+set load_file $plugins_path/emacs/emacs.load
+
```
bucaran: You could use `before_all` instead, and it should work too.

plugins/emacs/spec/emacs.spec.fish
```diff
@@ -0,0 +1,88 @@
+set spec_path (dirname (status --current-filename))
+set plugins_path (cd $spec_path; and cd ../..; and pwd)
```
bucaran: You can use `$fish_path/plugins` instead.

bucaran:
@oneKelvinSmith How's it going?

oneKelvinSmith:
I've submitted another pull request with your comments incorporated and some other cleanup.

I've also added two more alias functions `et` and `ec`. Please have a look at #408.

Thank you. 

p.s. I'll close this request.

PR #405, opened 2015-03-13
Add interactive mode to dpaste

RomanHargrave:
Allows `dpaste` to be run with the `-i` flag. Doing this will create a temporary file, open `editor`, and then pipe the result in to `__dpaste_send`. This is great for quickly editing and uploading arbitrary text input from the user to a paste site.

In order for this to work `__dpaste_send` now returns the exit status of cURL, which the entire plugin benefits from.

### DIFF
--- a/plugins/dpaste/dpaste.fish
+++ b/plugins/dpaste/dpaste.fish
@@ -26,6 +26,8 @@ function __dpaste_set_defaults
   set -g __dpaste_send_url (eval 'echo $__dpaste_url_'$suffix)
   set -q __dpaste_send_url; or set -g __dpaste_send_url $__dpaste_url_dpaste_de
   set -g __dpaste_eat_once 0
+
+  set -q __dpaste_editor; or set -g __dpaste_editor (readlink (which editor))
 end
 
 function __dpaste_send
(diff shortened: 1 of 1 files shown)


bucaran:
@RomanHargrave How would you feel about adding some tests?

bucaran:
```
dpaste -i
fish: Unknown command 'dpaste_interactive.QTP84'
/Users/b/.oh-my-fish/plugins/dpaste/dpaste.fish (line 104): begin; dpaste_interactive.QTP84 
                                                                   ^
in . (source) call of file '-',
	called on line 137 of file '/Users/b/.oh-my-fish/plugins/dpaste/dpaste.fish',

in function 'dpaste',
	called on standard input,
	with parameter list '-i'

No content given.
```

RomanHargrave:
My OS is Linux 4 rc 3 (Debian 8), editor is a command that usually comes with XDG, so I just assumed most would have it. In doing that I failed to account for Mac OS (if that's what you use). I'll add a function to comprehend the editor and add it.

RomanHargrave:
It looks like __dpaste _parse_expires is flattening $argv. I'll finish up later.

RomanHargrave:
Woops. That was not the comment button.

bucaran:
Hey @RomanHargrave!

Yes, we try to support both OSX and Linux, which is why then are new plugins and / or updates, we would like to use [`fish-spec`](https://github.com/bpinto/oh-my-fish/tree/master/plugins/fish-spec) to test stuff.

Anyhow, I will be closing this PR for now. If you can fix the issue please by all means do reopen. Also, don't forget to always pull rebase before pushing your updates.

PR #411, opened 2015-03-24
Adding plugin for exenv.

janlindblom:
Exenv is an Elixir version manager in the spirit of rbenv, ndenv, plenv, pyenv etc.

### DIFF
--- a/plugins/exenv/exenv.load
+++ b/plugins/exenv/exenv.load
@@ -1,0 +1,7 @@
+if test -n "$EXENV_ROOT"
+  _prepend_path $EXENV_ROOT/bin
+  _prepend_path $EXENV_ROOT/shims
+else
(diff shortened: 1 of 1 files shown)


bucaran:
Hey @janlindblom I haven't checked `exenv`, but if it is an exact clone of `rbenv` / `pyenv`, then that's probably not quite complete as we are probably missing:

```
if status --is-interactive
  # Make sure to symlink path/to/xxenv to with $PATH.
  . (exenv init - | psub)
end
```

I haven't got around this, but I am thinking instead of crafting a `xenv` plugin that would do this for all `rbenv`, `pyenv`, `ndenv`, `plenv`, etc.

bucaran:
Closing due to to activity.

PR #422, opened 2015-04-07
Port oh-my-zsh git plugin to oh-my-fish.

KenPowers:
This is a port of the latest [oh-my-zsh git plugin](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/git/git.plugin.zsh) to oh-my-fish. Pretty self explanatory.


bpinto:
Hi @KenPowers thanks for this, but alias are personal and we think alias are not to be included inside omf

bucaran:
Closely related #241

PR #423, opened 2015-04-11
adds the trout theme

phaedryx:
The prompt that I use. I thought I'd share.


bpinto:
That's cool! Thanks for sharing!

PR #424, opened 2015-04-12
Fix duplicated virtualenv prompt on Clearance theme

slapresta:
When using the Clearance theme inside an activated virtualenv, both the theme and the virtualenv activation script show a prompt indicating the name of the virtualenv prompt.

This pull request disables virtualenv's prompt, leaving only the theme's.

### DIFF
--- a/themes/clearance/fish_prompt.fish
+++ b/themes/clearance/fish_prompt.fish
@@ -31,6 +31,7 @@ function fish_prompt
   echo -e ''
 
   # Display [venvname] if in a virtualenv
+  set -gx VIRTUAL_ENV_DISABLE_PROMPT 1
   if set -q VIRTUAL_ENV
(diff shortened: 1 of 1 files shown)


bobthecow:
I'm hesitant to override non-OMF environment variables inside themes. I'd rather add something to the README to say "You might get a double-virtualenv prompt if you use this theme. Either disable the virtualenv prompt by setting `...`, or use virtualfish instead."

bucaran:
@bobthecow :+1:

bpinto:
ping @slapresta

slapresta:
Sorry, I kinda forgot about this for a while. Thanks for pinging me!

Adding info to the README is indeed one option. Another option would be to remove the theme's supplemental prompt altogether and let virtualenv display its native one; truth is, I don't really understand why the theme provides a virtualenv prompt.

bpinto:
As we have split the repo, you might want to add this here instead: https://github.com/oh-my-fish/theme-clearance

PR #435, opened 2015-04-24
Add grc plugin, budspencer theme update

tannhuber:
grc plugin added

### DIFF
--- a/plugins/grc/grc.load
+++ b/plugins/grc/grc.load
@@ -1,0 +1,39 @@
+# These methods override the default calls to append `grc exec`.
+# Please send a message if you find another command that can be
+# used with `grc`.
+set -l execs cat           \
+             configure     \
+             dig           \
+             gcc           \
+             g++           \
+             ls            \
+             make          \
+             netstat       \
+             ping          \
+             ps            \
+             tail          \
+             traceroute
+
+if set -q grc_plugin_execs
+  set execs $execs $grc_plugin_execs
+end
+
+# Fish 2.1.1+ has support for --inherit-variable
+set -l do_eval (echo $FISH_VERSION | grep 2.1.1-)
+
+for executable in $execs
+  if test -z "$do_eval"
+    eval "function $executable; __execute_as_grc $executable \$argv; end"
+  else
+    function $executable --inherit-variable executable
+      __execute_as_grc $executable $argv
+    end
+  end
+end
+
+function __execute_as_grc
+  if [ $argv[1] = 'ls' ]
+    set argv $argv '--color'
+  end
+  command grc -es --colour=auto $argv
+end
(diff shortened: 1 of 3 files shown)


plugins/grc/grc.load
```diff
@@ -0,0 +1,38 @@
...
+             ps            \
+             tail          \
+             traceroute
+
+if set -q grc_plugin_execs
+  set execs $execs $grc_plugin_execs
+end
+
+# Fish 2.1.1+ has support for --inherit-variable
+set -l do_eval (echo $FISH_VERSION | grep 2.1.1-)
```
bpinto: This grep does not work for fish 2.1.2. =/

themes/budspencer/fish_prompt.fish
```diff
@@ -509,6 +509,7 @@ function __budspencer_edit_commandline -d 'Open current commandline with your ed
     commandline ''
   end
   rm $budspencer_tmpfile
+  __budspencer_preexec
```
bpinto: What is this fixing?

themes/budspencer/fish_prompt.fish
```diff
@@ -516,9 +517,9 @@ end
 ################
 function __budspencer_prompt_git_branch -d 'Return the current branch name'
   set -l branch (command git symbolic-ref HEAD ^ /dev/null | sed -e 's|^refs/heads/||')
-  if not test $branch > /dev/null
+  if [ "x$branch" = "x" ]
```
bpinto: I don't like this hack, do you find it better to read than the previous code?

bucaran: You mean style, because this is not a hack. I don't like either though.

bpinto: It's a hack imo.

bucaran: Oh, I see, you are not talking about `[`, but `"x$branch" = "x" `.

Hah, that's a hack. I have no comment about that though. 

My suggestion is to use `if test` instead of `[`, that's it.

bpinto: Oh I didn't even notice the `test`, but I agree with you!

bucaran: :laughing:

plugins/grc/grc.load
tannhuber: That's the reason why it's there, isn't it? I copied these lines from the bundler plugin that you suggested.

themes/budspencer/fish_prompt.fish
tannhuber: It executes the commandline automatically after editing the commandline with $EDITOR. I don't need to press `Return` key anymore.

themes/budspencer/fish_prompt.fish
tannhuber: I can reverse it. Although I don't like the pipe to `/dev/null`.

bpinto:
@tannhuber Could you add it to https://github.com/oh-my-fish/plugin-grc ?

PR #436, opened 2015-04-24
[bobthefish] allow git untracked files override

### LINKED ISSUE #434: bobthefish: git ls-files taking forever
The `git ls-files` command in `__bobthefish_prompt_git` is taking forever (minutes) on large NFS mounts.

Adding `--directory` removes that delay, but I'm unsure if functionality is affected.

bobthecow:
Showing untracked files can be (very) slow on large NFS mounts. Allow the user to disable the '...' output when there are untracked files, either globally via fish config:

    set -g theme_display_git_untracked no

 or locally via git repo config:

    git config bash.showUntrackedFiles false

Fixes #434

### DIFF
--- a/themes/bobthefish/fish_prompt.fish
+++ b/themes/bobthefish/fish_prompt.fish
@@ -13,6 +13,7 @@
 # You can override some default options in your config.fish:
 #
 #     set -g theme_display_git no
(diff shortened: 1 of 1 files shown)


bpinto:
Nice git config!

PR #437, opened 2015-04-26
Some improvements on robbyrussel theme

kiRach:
1. The first improvement is added additional space between arrow and directory name:
    Before:
    Before
    After:
    After

2. The second one is about additional option to configure git to not show untracked files as dirty in status. Original oh-my-zsh has this option and I missed it in oh-my-fish, so I've added this option.

### DIFF
--- a/themes/robbyrussell/fish_prompt.fish
+++ b/themes/robbyrussell/fish_prompt.fish
@@ -1,18 +1,19 @@
 # name: RobbyRussel
 #
 # You can override some default options in your config.fish:
-#   set -g theme_disable_untracked_files_dirty yes
+#   set -g theme_display_git_untracked no
 
 function _git_branch_name
   echo (command git symbolic-ref HEAD ^/dev/null | sed -e 's|^refs/heads/||')
 end
 
 function _is_git_dirty
-  if [ "$theme_disable_untracked_files_dirty" = 'yes' ]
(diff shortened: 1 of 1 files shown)


themes/robbyrussell/fish_prompt.fish
```diff
@@ -1,10 +1,18 @@
...
   echo (command git symbolic-ref HEAD ^/dev/null | sed -e 's|^refs/heads/||')
 end
 
 function _is_git_dirty
-  echo (command git status -s --ignore-submodules=dirty ^/dev/null)
+  if [ "$theme_disable_untracked_files_dirty" = 'yes' ]
+    echo (command git status -s --ignore-submodules=dirty --untracked-files=no ^/dev/null)
+  else
+    echo (command git status -s --ignore-submodules=dirty ^/dev/null)
+  end
```
bobthecow: Use `theme_display_git_untracked` to match what bobthefish does?

https://github.com/bpinto/oh-my-fish/blob/master/themes/bobthefish/fish_prompt.fish#L16

Also, consider supporting `bash.showUntrackedFiles` git config as well:

https://github.com/bpinto/oh-my-fish/blob/master/themes/bobthefish/fish_prompt.fish#L290-L295

kiRach:
@bobthecow Thanks for your comment! Added changes according your recommendations.

bobthecow:
@kiRach no problem :)

bucaran:
@kiRach Thanks!

PR #440, opened 2015-04-29
append to config.fish instead of overwriting

### LINKED ISSUE #439: Installation overwrote my config.fish
I had a config.fish file with a custom PATH among other things, and when I installed oh-my-fish, that was completely overwriten! :(

josiahsprague:
This is my proposed change. Fixes #439

if there is an existing config.fish, append to the end of it instead of overwriting.


bpinto:
Appending may cause unknown errors to happen because these settings could affect the framework load. It's not recommended to add personal configurations on this file, but to use custom plugins and load files.

PR #441, opened 2015-04-29
Added support for svn in prompt

stieglma:
This changes the fish_prompt in a way that the prompt in svn working copys looks like in a git checkout (branchname and revision is shown)


bpinto:
@stieglma Could you open this PR on https://github.com/oh-my-fish/theme-agnoster ?

PR #446, opened 2015-05-05
Fix fry plugin

sschepens:


bucaran:
I don't use the plugin so I am not exactly sure how this _fixes_ the plugin, but below is a list of some of the flags for `test`.
```

       o -d FILE returns true if FILE is a directory.

       o -e FILE returns true if FILE exists.

       o -f FILE returns true if FILE is a regular file.

```

sschepens:
Fry plugin didn't even make sense, it tested that a file was a directory, if that was true then it would source the file...
I changed to test for files and added an alternative path to the file.

bpinto:
@sschepens Could you open this PR on https://github.com/oh-my-fish/plugin-fry ?

PR #447, opened 2015-05-05
FIxing a problem caused by the strange "command command -v rbenv" command

Bastes:
(see discussion on commit f4faa2e812c96c371fabee46d1c02e8cf2b3d6b2)

The fix inverts the position of the strange "command command -v rbenv" that fails noisily, so that the standard position for the rbenv file is evaluated first and the failing line doesn't need to be triggered in these cases.

Tested on my environment, it seems to work.

### DIFF
--- a/plugins/rbenv/rbenv.load
+++ b/plugins/rbenv/rbenv.load
@@ -6,11 +6,11 @@
 if [ -e "$RBENV_BIN_ROOT/rbenv" ]
   set rbenv_binary "$RBENV_BIN_ROOT/rbenv"
   _prepend_path "$RBENV_BIN_ROOT"
-else if [ (command command -v rbenv) ]
-  set rbenv_binary (command command -v rbenv)
 else if [ -e "$HOME/.rbenv/bin/rbenv" ]
   set rbenv_binary "$HOME/.rbenv/bin/rbenv"
   _prepend_path "$HOME/.rbenv/bin"
+else if [ (command command -v rbenv) ]
+  set rbenv_binary (command command -v rbenv)
 else
   echo "Could not find rbenv. Make sure it's on your system path, in your home directory or set the RBENV_BIN_ROOT environment variable pointing to the directory where you unpacked rbenv."
   exit 1
--- a/plugins/rbenv/rbenv.load
+++ b/plugins/rbenv/rbenv.load
@@ -6,11 +6,11 @@
 if [ -e "$RBENV_BIN_ROOT/rbenv" ]
   set rbenv_binary "$RBENV_BIN_ROOT/rbenv"
   _prepend_path "$RBENV_BIN_ROOT"
+else if [ (which command; and command command -v rbenv) ]
+  set rbenv_binary (command command -v rbenv)
 else if [ -e "$HOME/.rbenv/bin/rbenv" ]
   set rbenv_binary "$HOME/.rbenv/bin/rbenv"
   _prepend_path "$HOME/.rbenv/bin"
-else if [ (command command -v rbenv) ]
-  set rbenv_binary (command command -v rbenv)
 else
   echo "Could not find rbenv. Make sure it's on your system path, in your home directory or set the RBENV_BIN_ROOT environment variable pointing to the directory where you unpacked rbenv."
   exit 1


bobthecow:
Rather than punting and hiding the symptom, let's try to fix the error :)

What happens when you redirect stderr on your system?

    command command -v rbenv ^/dev/null; or echo "failed"

Bastes:
Hey @bobthecow ^^ thanks for your quick reply.

I tried what you said, It says:
```
command : commande introuvable
command command -v rbenv ^/dev/null; or echo "failed"
        ^
failed
```
(my Ubuntu's in french, "command introuvable" means "unknown command")

Note that the "command" command exists on my computer, and that I can call it with whatever other command (it runs the command itself).

```
$ command --help

       command - command - run a program


command - run a program
   Synopsis
       command COMMANDNAME [OPTIONS...]

   Description
       command forces the shell to execute the program COMMANDNAME and ignore
       any functions or builtins with the same name.

   Example
       command ls causes fish to execute the ls program, even if an 'ls'
       function exists.
$ command rbenv -v
rbenv 0.4.0-143-g4d72eef
```

On the discussion I mentioned in the pull request, we scratched our heads without understanding what "command command" call was trying to achieve, it seems strange to use a command that calls another command  to call itself (supposing it is the same "command" on Ubuntu and on whatever system this part of the plugin was designed for).

Then again, I'm not much of a fish scripter, so feel free to enlighten me on whatever purpose you think it might have ^^

bobthecow:
Got it. On BSD-ish systems (e.g. OS X) there's a bin called `/usr/bin/command`. This is trying to use `command` (the fish builtin) to execute `command` (the script located at `/usr/bin/command`) and failing, because you've only got the builtin.

Bastes:
Then I guess this fix still makes sense: the error is not relevant to a system that doesn't have this special brand of ```/usr/bin/command``` ; if there's no way to silence it, people in my situation, that have installed their rbenv in their home, don't get any benefit from having the ```command command``` test executed before the one that actually locates the command, but do get frightened to see an error that doesn't concern them anyways. The other people, that have installed it somewhere else and would benefit from using this ```command command``` test will still get there and have the right command located then.

Anyways, I don't want to be pushy about it, I know managing a repository of a lot of various plugins can be tricky and I wouldn't pretend to understand exactly everything this plugin does, so take what you will of this, and in the meantime I have my fix home that works for me ^^

Thanks for the chat, really appreciate how fast and nicely you reacted to my little initiative.

bobthecow:
but moving it around still means the issue is there. you just hid it for *you* :)

we can fix the actual cause rather than hiding the symptom. just check for `which command` before trying `command command ...`.

Bastes:
Ok, this seems to work (on my system at least) ; I'm committing a new pull request with the proper fix. Thanks for your help ^^

Bastes:
So, how about merging now? ^^

bpinto:
@Bastes This looks good, could you rebase it?

Bastes:
Sure, right away ^^

Bastes:
There, all on top of master ^^

bpinto:
:+1:

bpinto:
Thanks for your work on this!

Bastes:
@bobthecow did most of the smarts ^^ but thanks.

wesbaker:
I *think* this is causing some issues for me on OS X, specifically:

> Could not find rbenv. Make sure it's on your system path, in your home directory or set the RBENV_BIN_ROOT environment variable pointing to the directory where you unpacked rbenv.

I don't have `RBENV_BIN_ROOT` defined, but `which command` and `command command -v rbenv` both return something:

```sh
➜  ~ which command
/usr/bin/command
➜  ~ command command -v rbenv
/usr/local/bin/rbenv
➜  ~ which command; and command command -v rbenv
/usr/bin/command
/usr/local/bin/rbenv
```

Looking at the if statement, everything should work. When I get rid of `which command; and` everything works as expected. Any ideas?

wesbaker:
Also, am I correct in assuming that this should run `rbenv rehash` after doing something like a `gem install` or am I off base?

bpinto:
I could be wrong, but I don't think `rbenv rehash` should be automatically installed, there is another gem for that if you are insterested: https://github.com/sstephenson/rbenv-gem-rehash

bpinto:
Regarding the `command command` issue, I am not sure if I understood what is happening.

bucaran:
`command command` is terrible code, but in terms of what's happening is just making sure `fish` builtin `command` gets called as opposed to any user defined `command`. Crazy.

wesbaker:
Could this be simplified to something like this:

```sh
else if [ (which rbenv) ]                                                       
  set rbenv_binary (which rbenv)
```

bucaran:
Check out the [same plugin](https://github.com/hauleth/wa-rbenv) for Wahoo. Very simple, basically:

```fish
  set -x RBENV_SHELL fish
  if status --is-interactive
    source (rbenv init - | psub)
  end
```

bpinto:
@wesbaker  It's likely that it could be simplified. Want to try to open a PR?

wesbaker:
I don't mind trying, it'd basically be what I have above. @bucaran do you have a specific suggestion for this plugin?

bucaran:
@wesbaker Yes, it should be rewritten by someone who uses `rbenv` a lot.

PR #448, opened 2015-05-06
Dangerous theme

tannhuber:
I found some spare time to translate my dangerous theme from zsh to fish. It would be the first oh-my-fish theme starting with 'd'. :-)

The functionality is similar to budspencer.
![screenshot][screenshot]
[screenshot]: https://raw.githubusercontent.com/tannhuber/media/master/dangerous.gif

### DIFF
--- a/themes/dangerous/README.md
+++ b/themes/dangerous/README.md
@@ -1,0 +1,223 @@
+# Dangerous theme
+
+A theme for the 70s, makes fish even more powerful. It's equipped with a hell
+of a lot of nice functions and key bindings to speed up your workflow.
+
+The philosophy behind dangerous theme:
+
+* Delimit the prompt lines clearly to the output.
+* Produce output only if it contains relevant information.
+* Show useful notifications within the prompt.
+* Equip the prompt with toggle commands in order to change its style
+  interactively.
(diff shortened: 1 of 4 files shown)


bucaran:
Hey @tannhuber! I love budspencer, but how's this theme different from it? I know it's not affecting other parts of the code, but there's technical debt associated with any addition.

bobthecow:
@tannhuber What are you using for the colorized `ls` output?

tannhuber:
@bucaran Yes, there're only few changes related to budspencer: 
* It looks differently.
* No powerline font is needed.
* It's possible to hide **all** the segments, not only the PWD segment. (Perhaps I'll change that behaviour in budspencer theme as well. At the moment I'm playing with both alternatives.)

I originally wrote three identical behaving but differently looking themes for zsh, one as default theme, the second one for root shells and the last one for ssh connections. For me it's important seeing at a glance where I am. Besides, I like some variety. The dangerous theme was for root shells. That's the reason why I called it dangerous.

@bobthecow Look here: #435

bpinto:
@tannhuber would you push it here instead? https://github.com/oh-my-fish/theme-dangerous

PR #453, opened 2015-05-13
Add git functions

ubermonkey:
Bunch of git shortcuts for fish shell

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -1,0 +1,20 @@
+Bunch of git shortcuts for fish shell. [Author][1].
+*All commands accept arguments as usual.*
+
+`gi` —  `git init`
+
+`ga` — `git add`
+
+`gs` — `git status`
+
+`gr` — `git rm`
+
+`gc` — `git commit`
+
+`gcl` — `git clone`
+
+`gpll` — `git pull`
+
+`gpsh` — `git push`
+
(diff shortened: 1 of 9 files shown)


bucaran:
@ubermonkey Thanks! But I think git shortcuts are better left off to the user. I personally prefer using a `.gitconfig`. [For example](https://gist.github.com/pksunkara/988716).

ubermonkey:
@bucaran ok. I just thought that user should be able to enable `git` plugin in the `config.fish`. 'Cause I prefer to control everything about shell commands from shell config, and personally I missed git plugin.

bobthecow:
You absolutely can put it in your shell, but as @bucaran says, shortcuts are a personal thing, so we've avoided adding them to the global project. oh-my-fish supports personal plugins: just drop this `git` folder in `.config/oh-my-fish/plugins/` and it'll work the same as if it were in the global oh-my-fish folder.

bucaran:
@ubermonkey What @bobthecow said! :smile:

ubermonkey:
I've took a look on other plugins, that what you say make sense, now I agree that 'bunch of shortcuts' is not enough to be a plugin. Sorry for disturbing.

bucaran:
@ubermonkey It's cool dude. Worry not.

PR #456, opened 2015-05-17
Adding newline prompt option to Agnoster

daogilvie:
This is a fish port of the pending pull-request into oh-my-zsh discussed in this thread: 
https://github.com/robbyrussell/oh-my-zsh/pull/2984 by the user wenlock.
Adds ability to set env variable PROMPT_ON_NEWLINE = true.
Setting this causes prompt to display as indicated in attached image.
Image also shows results of tests before and after to demonstrate no breaking change.
screenshot

### DIFF
--- a/themes/agnoster/fish_prompt.fish
+++ b/themes/agnoster/fish_prompt.fish
@@ -146,6 +146,7 @@ end
 # ===========================
 
 function fish_prompt
+  set -g RETVAL $status
   if eval $PROMPT_ON_NEWLINE == "true"
     echo -n "╭─"
   end
--- a/themes/agnoster/fish_prompt.fish
+++ b/themes/agnoster/fish_prompt.fish
@@ -147,17 +147,12 @@ end
 
 function fish_prompt
   set -g RETVAL $status
-  if eval $PROMPT_ON_NEWLINE == "true"
-    echo -n "╭─"
-  end
-
-  prompt_status
-  prompt_user
-  prompt_dir
-  prompt_git
-  prompt_finish
-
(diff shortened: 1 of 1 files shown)


daogilvie:
Have just discovered incompatible with python virtualenv. Need to fix. Closing.

bucaran:
@daogilvie I'll keep an eye on this. Thanks @daogilvie

daogilvie:
No problem @bucaran ; I'm not actually sure how I would fix it against venv's default behaviour. I'm stopping using venv as much these days in favour of containers and full vms. Maybe I'll think of something along the way, but like I say I'm not sure how right now.

daogilvie:
Ok @bucaran, I'm not sure how to fix this. Given that it's a problem mainly with venv, I'll reopen this request, and you can decide whether or not you want it in.

bpinto:
@daogilvie what is the incompatibility issue?

daogilvie:
@bpinto When you activate a venv the default scripts just munge the venv's name into the start of the prompt, and their use of printf seems to negate the line splitting. So you end up with everything back on one line, but the second line start character is printed after the final prompt segment. Looks like this:
screen shot 2015-05-31 at 15 56 54

daogilvie:
I don't know how I would fix it. If it's a problem, please feel free to reject the pull request. When not using the newline option there is no issue.

bpinto:
I honestly don't see much of an issue as it only happens when the newline options is being used. Could you reopen this on https://github.com/oh-my-fish/theme-agnoster ?

PR #462, opened 2015-05-26
Vundle style plugins (external plugins)

bpinto:
- [x] Update documentation.
- [x] Upgrade guide.
- [x] Support update.
- [x] Add documentation to the code.
- [ ] Add tests.


functions/fish.fish
```diff
@@ -0,0 +1,20 @@
+# NAME
+#   fish - Extend default fish binary
+#
+# DESCRIPTION
+#   Extend fish binary to support plugins and themes installation
+#
+function fish -d "Extend fish binary"
```
rominf: I don't like this. How can I run `install` file with `fish`?
```
fish.packages --install
```
looks quite good for me.

bobthecow: Agreed. I'm not a fan of changing what `fish` means :)

bobthecow: Why not have an `omf` binary to handle all the oh-my-fish things?

bobthecow:
We could remove the dependency on git by using github's tarball downloads. Not sure if that's worth it, though.

bobthecow:
Install and update seem to depend on `$fish_theme` and `$fish_plugins` being set, but those are deprecated, right?

functions/fish.fish
bpinto: Okay!

bpinto:
We could remove git dependency, although I like how it makes things simple. I think I'm not opposed to supporting tarballs, just don't think there are more interesting features to add first.

`$fish_theme` and `$fish_plugins` and deprecated to be set manually but we use them by install and update so that the system continues working the same, meaning we don't break any functionality when someone upgrades omf.

PR #465, opened 2015-05-30
Fix broken fish-spec link in CONTRIB~.md

lloydde:
Also
- Now the one inline link is also footnote.
- Consistent no hard wrap here and in README.md


bpinto:
@lloydde Hey thanks for that! The more well explained the README and the wiki pages are the better! I am not good on these though... ;(

PR #467, opened 2015-05-31
Create fedora plugin

### LINKED ISSUE #466: Add Fedora plugin
I ported my favorite Fedora plugin for zsh to fish. It functions similarly to the exiting archlinux plugin in that it just provides simple package manager aliases. The source repo doesn't have a license, but I credited the author in the README, so hopefully that should suffice.

[Link to the fish plugin in my fork](https://github.com/Poorchop/oh-my-fish/tree/fedora-plugin/plugins/fedora)

Poorchop:
Closes #466

### DIFF
--- a/plugins/fedora/README.md
+++ b/plugins/fedora/README.md
@@ -1,0 +1,17 @@
+## Fedora
+
+Port of TingPing's [fedora.plugin.zsh](https://github.com/TingPing/zsh/blob/master/plugins/fedora/fedora.plugin.zsh)
+
+#### commands (compatible with both yum and dnf)
+yse  = $pkgmgr search  
+ypro = $pkgmgr provides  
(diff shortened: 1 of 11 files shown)


bpinto:
This is interesting...

We do not add plugins with aliases to this repository because aliases are personal.

But now, instead of bundling all plugins and themes inside this single repo, we are now creating them on individual repositories at `https://github.com/oh-my-fish`. With this change, it's going to be harder to maintain this rule.

ping @bobthecow @bucaran

bucaran:
Actually, I still argue that aliases are a no-no even in a decentralized repo, so I don't allow them on Wahoo, even though everyone is responsible for their own package.

bpinto:
I am inclined to do the same. Thanks for the PR though @Poorchop

Issue #474, opened 2015-06-05
omf self-update changes current working directory


derekstavis:
Using self-update changes the current working directory to oh-my-fish install directory.

```
➜  ~ 
➜  ~ omf self-update
Oh My Fish has been successfully updated.
➜  .oh-my-fish git:(master) pwd
/Users/derek/.oh-my-fish
➜  .oh-my-fish git:(master) 
```

jonscottclark:
Could perhaps use **pushd** instead of **cd** and then add a **popd** at the end (ensuring popd output is silenced)? https://github.com/oh-my-fish/oh-my-fish/blob/master/plugins/omf/omf.git.fish#L10

derekstavis:
I've made this patch using @jonscottclark suggestion and it solves the problem here.
Although, I'm not sure if it follows project's coding style.

bpinto:
Thanks for the PR @derekstavis and for the idea @jonscottclark

PR #475, opened 2015-06-05
fix issue #474 by using pushd/popd

derekstavis:


bpinto:
Thanks for the improvement!

PR #499, opened 2015-07-20
fixes #496 Do not throw warnings when theme is not set

### LINKED ISSUE #496: "omf list" displays error when not using a theme
I use OMF for its plugins but do not use the theme functionality. When running the "omf list" command, the plugins display correctly but after the "Theme:" heading, it displays an error.

Example session:

`~ $ omf list`

```
Plugins: 
z gi git-flow grc mc msg

Theme: 
Array index out of bounds
~/.oh-my-fish/plugins/omf/omf.log.fish (line 5):       echo -e (set_color $argv[1])$argv[2..-1](set_color normal)
                                                                                        ^
in function “omf.log”
	called on line 49 of file ~/.oh-my-fish/plugins/omf/omf.packages.fish
	with parameter list “normal”

in function “omf.packages”
	called on line 15 of file ~/.oh-my-fish/plugins/omf/omf.fish
	with parameter list “--list”

in function “omf”
	called on standard input
	with parameter list “list”
```

rholak:
Check for existence of $fish_plugins and $fish_theme when using the "omf list" feature. This prevents errors/warnings from being displayed when they are not necessary in this case.

### DIFF
--- a/plugins/omf/omf.packages.fish
+++ b/plugins/omf/omf.packages.fish
@@ -42,11 +42,15 @@ function omf.packages --argument-names options -d 'Manage all plugins and themes
       end
     case '--list'
       omf.log yellow 'Plugins: '
-      omf.log normal $fish_plugins
+      if test -n "$fish_plugins"
+        omf.log normal $fish_plugins
+      end
 
       omf.log normal ''
       omf.log yellow 'Theme: '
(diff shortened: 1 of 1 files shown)


bpinto:
Thanks @rholak !! :+1:

PR #510, opened 2015-08-26
Only show deprecation warnings when omf is loaded.

ohspite:
The oh-my-fish.fish startup file is sourced when changing themes after
the `fish_theme` variable is set.  The deprecation warning should only
be shown if the user sets the deprecated variables.


bpinto:
Hey, I like this change!
As you are opening a couple PRs, you might be interested on https://github.com/oh-my-fish/oh-my-fish/issues/507

ohspite:
Oh cool, thanks for letting me know.  I had not seen that.

PR #511, opened 2015-08-26
Fix error handling when github API is unavailable.

ohspite:
The `exit` (instead of `return`) caused the shell to quit if there was
an error with github.


bpinto:
Thanks for fixing it!

PR #514, opened 2015-08-26
Rename `omf use` to `omf theme`.

bpinto:
The first time I saw `use`, I thought I could use the command for
everything, including packages, but it's only used for themes. So, it's
better to have a clearer name.


bucaran:
:+1:

PR #515, opened 2015-08-26
Current in use theme should not be deletable

bpinto:
Get default theme by reading from $OMG_CONFIG/theme file and skip it
from the list of installed themes.

There is a bug with `basename`, notice how 'led' is not displayed on
`basename` call:

```
omf:fix-omf-remove* λ ls $OMF_PATH/themes/
l   led
omf:fix-omf-remove* λ basename $OMF_PATH/themes/*
l
```


bucaran:
Fixed with improved `basename` in bf52f5d57a8598f2e0c67fe20c5bf3359618f5ae

PR #516, opened 2015-08-26
Current in use theme should not be deletable

bpinto:
Get default theme by reading from $OMG_CONFIG/theme file and skip it
from the list of installed themes.


bucaran:
This would let you remove the "default" theme instead.

bpinto:
Cleared, no need to loop anything.

bucaran:
> Cleared, no need to loop anything.

That would be nice. Unfortunately we have to :cry: , unless you come up with a better way to skip the `default` theme.

PR #528, opened 2015-08-27
Merge `$OMF_CONFIG` and `$OMF_CUSTOM`.

bobthecow:
Per conversation with @bpinto in Gitter.

There's no need for two separate directories. You don't have a `.git` and `.git-custom` folder, you just put your config in `.git` :)

The most straightforward interpretation of XDG basedir spec is that user configuration for omf would go in `~/.config/omf`, so let's put it there (by default). The only question is whether omf-generated config (i.e. the `theme` file) should go there as well. By analogy with git, programmatically generated config should probably be merged in with user config. This also makes it so when a user clones their dotfiles to a new machine, both kinds of settings come with it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ Remove a theme or package.
 
 Scaffold out a new package or theme.
 
-> This creates a new directory under `$OMF_CUSTOM/{pkg | themes}/` with a template.
+> This creates a new directory under `$OMF_CONFIG/{pkg | themes}/` with a template.
 
 ## `omf submit` _`pkg/<name>`_ _`[<url>]`_
 
@@ -135,9 +135,9 @@ Uninstall Oh My Fish. See [uninstall](#uninstall) for more information.
 
 ## Startup
 
-This script runs each time a new session begins, autoloading packages, themes and your _custom_ path (dotfiles) in that order.
+This script runs each time a new session begins, autoloading packages, themes and your _config_ path in that order.
 
-The _custom_ path (`$HOME/.dotfiles` by default) is defined by `$OMF_CUSTOM` in `$HOME/.config/fish/config.fish`. Modify this to load your own dotfiles if you have any.
+The _config_ path (`~/.config/omf` by default) is defined by `$OMF_CONFIG` in `~/.config/fish/config.fish`. Modify this to load your own configuration if you have any.
 
 ## Core Library
 
--- a/bin/install
+++ b/bin/install
@@ -3,7 +3,7 @@
 # USAGE
 #   #1: curl -L git.io/omf | sh
 #   #2: curl -L git.io/omf > install && chmod +x install && ./install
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -135,9 +135,9 @@ Uninstall Oh My Fish. See [uninstall](#uninstall) for more information.
 
 ## Startup
 
-This script runs each time a new session begins, autoloading packages, themes and your _custom_ path (dotfiles) in that order.
+This script runs each time a new session begins, autoloading packages, themes and your _config_ path (dotfiles) in that order.
```
bpinto: Remove `dotfiles` reference?

README.md
```diff
@@ -135,9 +135,9 @@ Uninstall Oh My Fish. See [uninstall](#uninstall) for more information.
...
 
-The _custom_ path (`$HOME/.dotfiles` by default) is defined by `$OMF_CUSTOM` in `$HOME/.config/fish/config.fish`. Modify this to load your own dotfiles if you have any.
+The _config_ path (`$HOME/.dotfiles` by default) is defined by `$OMF_CONFIG` in `$HOME/.config/fish/config.fish`. Modify this to load your own dotfiles if you have any.
```
bpinto: It's no longer in `~/.dotfiles`, right?

I'd rather use `~` instead of `$HOME` too.

bobthecow: oh, sorry, i missed one :)

bobthecow: fixed.

README.md
bobthecow: fixed.

bin/install
```diff
@@ -3,7 +3,7 @@
 # USAGE
 #   #1: curl -L git.io/omf | sh
 #   #2: curl -L git.io/omf > install && chmod +x install && ./install
-#   #3: OMF_CUSTOM=~/.dotfiles curl -L git.io/omf | sh
+#   #3: OMF_CONFIG=~/.omf curl -L git.io/omf | sh
```
bpinto: `~/.config/omf`

bpinto:
LGTM
ping @bucaran

bin/install
bobthecow: i figured that was an example of how to add a custom config directory, so i made it not match the default on purpose.

bpinto: Ah, okay; :+1:

bucaran:
What about user's dotfiles? I have a `~/.rc` directory I want to use for my custom configuration, etc. Where should I put that now?

bobthecow:
Yes, there is a misunderstanding. This does nothing like that :)

The variable for your `~/.rc` directory is now called `$OMF_CONFIG` instead of `$OMF_CUSTOM`, but it still works exactly the same way. Here's mine:

screenshot_2015-08-28_01_43_20

bucaran:
Got it. This is wonderful then.

bucaran:
:+1:

PR #529, opened 2015-08-27
No refresh on install or theme

cap10morgan:
This should fix the execution halting problem in `omf update` because it leaves refreshing up to the caller of `omf_install_package` and `omf_theme`.


bpinto:
:+1:

bucaran:
> This should fix the execution halting problem...

What is this problem?

I still want to refresh after a package is installed so that it becomes available inmediately.

bpinto:
That still happens it's only fixing a bug.

Omf update was not updating themes as the execution was being halted by the many refresh calls.

bucaran:
For each package, I see. 
Good call :+1:

PR #530, opened 2015-08-27
Use fish's colors

bpinto:
Instead of using custom colors, reuse fish's colors.
Original fish colors here:
https://github.com/fish-shell/oh-my-fish/blob/4d628d5f43dbe8b138605d1f6ecbd762187afd7f/functions/restore_original_fish_colors.fish


bobthecow:
:+1:

bucaran:
:gun: :+1:

PR #531, opened 2015-08-27
Naming functions with `omf.` namespace.

bpinto:
Having a clear namespace `omf.` improves the readability of the code as
we clarify what is the function name and what is the namespace.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,7 +59,7 @@ The following syntax is more concise, but arguably less transparent.
 > You still may use `and` / `or` statements if you consider `if..else..then` to be overkill.
 
 ```fish
-set -q VAR; set -g VAR 42
+set -q VAR; or set -g VAR 42
 ```
 
 ### Functions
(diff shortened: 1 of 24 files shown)


CONTRIBUTING.md
```diff
@@ -80,16 +80,9 @@ function greet -a message -d "Display a greeting message"
...
 
-If you are writing a function inside another function, prefix the inner one with the parent's name.
-
-```fish
-function parent
-  function parent_child
-  end
-end
-```
+`fish` does not have private functions, so in order to avoid polluting the global namespace, use double underscore before your function name. For example, if you are writing a `ninja` plugin using `__ninja.function_name`.
```
bobthecow: `__ninja.` or `ninja.`?

bpinto: For private functions `__ninja.` for functions of the package `ninja.`.

bobthecow: :+1:

bucaran: OK

bucaran:
:+1:

CONTRIBUTING.md
bpinto: What is the benefit?

bucaran: Because it adds them to the global scope. So, it makes parsing the global function list harder and we don't want to pollute fish, you know

bpinto: lol... if only we had a rule that `__` means private, you would only need to skip the functions beginning with that name. :D

bucaran: Actually that gives me an idea using a file naming convention. We could parse a file and extract function names. The source it, then delete all the functions.

PR #532, opened 2015-08-27
Extracting template to a file and improving it

bpinto:
It's hard to read the template as it was being created using `echo`
instead of being defined on a separate file. Also adding some comments
to improve the configuration file readability.

Depends on https://github.com/fish-shell/oh-my-fish/pull/528

### DIFF
--- a/bin/install
+++ b/bin/install
@@ -92,10 +92,8 @@ omf_install() {
   touch ${fish_config_file} >/dev/null 2>&1
   test ! -w ${fish_config_file} && die "Fish configuration file is not writable, aborting."
 
-
-  echo "set -g OMF_PATH $(echo "${OMF_PATH}" | sed -e "s|$HOME|\$HOME|")" > ${fish_config_file}
-  echo "set -g OMF_CONFIG $(echo "${OMF_CONFIG}" | sed -e "s|$HOME|\$HOME|")" >> ${fish_config_file}
-  echo "source \$OMF_PATH/init.fish" >> ${fish_config_file}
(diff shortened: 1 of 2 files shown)


bobthecow:
:-1: this breaks xdg support

bpinto:
How does it?

bobthecow:
It's dynamically writing those paths to the config file, based on XDG env variables. Your version hard codes them :)

bpinto:
I see... well, now I have to go through my usual `sed` limitations... how do to the `sed` on the file. :D

bpinto:
`sed` works differently on OS X and linux... that's why I always have difficulties :(

bpinto:
Redone.
ping @bobthecow

bobthecow:
:+1: lgtm

we need to merge my config change so this will work :)

bucaran:
:-1:

bobthecow:
I think you misunderstand the purpose of this pull request.

bpinto:
I agree :)

bpinto:
@bucaran 
Why did you close? What is the problem? Let's discuss, not close the PR.

bucaran:
@bpinto This got closed when I merged @bobthecow PR by pulling directly from his copy and adding my changes on top. A lot of stuff happened last night.

So, what are we trying to accomplish with this?

bpinto:
Just removing hard to read `echo` calls from the init file and moving them to a template file.

bucaran:
Cool. This is not about editing the config file, just having config templates. 

LGTM.

PR #539, opened 2015-08-28
Change theme when creating a new theme

bpinto:
We already change the current directory to the theme dir, to make it
easier for development, we also change the current theme.


pkg/omf/cli/omf.new.fish
```diff
@@ -27,6 +27,8 @@ function omf.new -a option name
       $github $user $name
 
     echo (omf::em)"Switched to $dir"(omf::off)
+
+    test "$option" = themes; and omf.theme $name; and refresh
```
bucaran: Better `if` here.

bucaran:
Good.

bpinto:
@bucaran updated.

Issue #517, opened 2015-08-26
Create a oh-my-fish theme


bpinto:
We should have a default cool theme for omf. Using our logo on the theme would be a big plus.

bobthecow:
```
ln -s ./bobthefish ./omf
```

;)

bucaran:
@bpinto Seriously, why not? We should all pair-review the code, find bugs, improve whatever we can, improve comments.

@bobthecow What do you think about changing the name to just `bob` or `default` or `omf`?

First we need to make sure to download the powerline patched fonts, some like:

```sh
  fonts_dir="$HOME/Library/Fonts"
  powerline=$(mktemp -d $HOME/XXXX)
  git clone "https://github.com/powerline/fonts" $powerline
  find $powerline -name "*.[o,t]tf" | while read font
    do
      cp "$font" "$fonts_dir"
    done
  rm -rf $powerline
```

bpinto:
I know many (everyone? :D) use `bobthefish`, but I don't know if it should be the default.

The problem that I have with it is that it is not blazing fast as it has a lot of information to display, and the reason why I moved to fish was for its speed.

I'd rather have a very simple theme as the default.

bpinto:
Download power line fonts? I can see that as a plugin, but not as part of omf's bootstrap.

bucaran:
> I'd rather have a very simple theme as the default.

I can understand that. Should it be powerline style or plain?

bpinto:
Is there a way to check if a font is patched?

bucaran:
We can check whether fonts are installed.

bobthecow:
But we can't tell whether a shell is currently using it :(

bucaran:
In that case, we need a `fonts` plugin that does nothing, but to download the fonts if they are not already installed.

bucaran:
@bpinto How should the theme look like?

One more thing. @bobthecow Can you go over: https://github.com/fish-shell/oh-my-fish/tree/master/lib/git and see if everything is alright?

bpinto:
@bucaran A plugin to download the font is not bad... many have trouble with this.
Regarding the default theme, I don't know yet. I am bad with design :D

tannhuber:
I think the default theme is most important for new fish users. Experienced people probably use a different theme. So there's no need to implement complex functionality that is not obvious for former bash or zsh users. However, I think it's a good idea to use impressive colors in order to show: »Oh-my-fish can if you want.«

[...]

PR #542, opened 2015-08-28
Reference #517 Add default theme to database

bpinto:


bpinto:
@bucaran I think we can fetch this theme while installing OMF, but no need to stop people from deleting it, I guess.

bucaran:
@bpinto Or we could revert to the original behavior when there was a default theme. I would say let's prevent people from deleting it. And yes, we can fetch it during the first OMF init.

bpinto:
@bucaran it would also work, your call.

bucaran:
:+1: Alright.

PR #548, opened 2015-08-29
describe command

sheldon:
Hi

I was struggling to find an easy way to browse all the available packages and find out what they do, so I added an info command to try achieve this.

It pulls the descriptions from the repos on github, so is a little slow. Potentially we could add this info into the local db to make it super quick, but I didn't want to change formats without discussing first.

What do you think?

### DIFF
--- a/pkg/omf/omf.fish
+++ b/pkg/omf/omf.fish
@@ -72,6 +72,13 @@ function omf -d "Oh My Fish"
     case "l" "li" "lis" "lst" "list"
       omf.list_local_packages | column
 
+    case "d" "desc" "describe"
+      if test (count $argv) -eq 1
+        omf.describe
+      else
+        omf.describe $argv[2..-1]
+      end
+
     case "i" "install" "get"
       if test (count $argv) -eq 1
         omf.list_db_packages | column
--- a/pkg/omf/cli/omf.describe.fish
+++ b/pkg/omf/cli/omf.describe.fish
@@ -1,0 +1,17 @@
+function omf.describe -a name
+  if test (count $argv) -eq 0
+    for package in (omf.list_db_packages)
+      echo $package - (omf.describe $package)
+    end
+  else
+    set package_path $OMF_PATH/db/pkg/$name
+    if test -e $package_path
+      set url (cat $package_path)
+      set repo (basename (dirname $url))/(basename $url)
+      curl -s https://api.github.com/repos/$repo | ruby -rjson -e 'j = JSON.parse(ARGF.read); puts j["description"]'
+    else
+      echo (omf::err)"$name is not a valid pkg."(omf::off) 1^&2
+      return $OMF_INVALID_ARG
+    end
+  end
+end
--- a/pkg/omf/cli/omf.help.fish
+++ b/pkg/omf/cli/omf.help.fish
@@ -5,6 +5,7 @@ function omf.help
 
   "(omf::dim)"Actions"(omf::off)"
          "(omf::em)"l"(omf::off)"ist  List local packages.
+     "(omf::em)"d"(omf::off)"escribe  Get information about what packages do.
       "(omf::em)"i"(omf::off)"nstall  Install one or more packages.
         "(omf::em)"t"(omf::off)"heme  List / Use themes.
        "(omf::em)"r"(omf::off)"emove  Remove a theme or package.
--- a/pkg/omf/completions/omf.fish
+++ b/pkg/omf/completions/omf.fish
@@ -12,19 +12,23 @@ end
 
 complete --no-files -c omf -d "Oh My Fish"
 
-complete -c omf -n "__omf.opt_is q query"      -a (printf "%s " (set | awk '{ printf $1"\n"; }'))
-complete -c omf -n "__omf.opt_is r rm remove"  -a (printf "%s " (omf.list_local_packages) (omf.list_installed_themes))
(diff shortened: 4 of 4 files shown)


pkg/omf/cli/omf.info.fish
```diff
@@ -0,0 +1,16 @@
+function omf.info -a name
+  if test (count $argv) -eq 0
+    for package in (omf.list_db_packages)
+      echo $package - (omf.info $package)
+    end
+  else
+    set url (cat $OMF_PATH/db/pkg/$name)
+    set repo (basename (dirname $url))/(basename $url)
+    curl -s https://api.github.com/repos/$repo | ruby -rjson -e 'j = JSON.parse(ARGF.read); puts j["description"]'
+    or begin
```
bucaran: Let's go with...
```fish
if not curl...
...
end
```

pkg/omf/cli/omf.info.fish
```diff
@@ -0,0 +1,16 @@
...
+      echo (omf::err)"$local_name is not a valid pkg."(omf::off) 1^&2
+      return $OMF_INVALID_ARG
+    end
+  end
+  return 0
```
bucaran: Is this return really necessary?

pkg/omf/omf.fish
```diff
@@ -72,6 +72,13 @@ function omf -d "Oh My Fish"
     case "l" "li" "lis" "lst" "list"
       omf.list_local_packages | column
 
+    case "info"
```
bucaran: The `i` in `info` collides with `install`. Let's find another name?

bucaran:
LGTM.

> It pulls the descriptions from the repos on github, so is a little slow. Potentially we could add this info into the local db to make it super quick, but I didn't want to change formats without discussing first.

I decided against this early on because it would increase complexity significantly to parse the registry. It's a simple text file.

Another thing is, we should remember that OMF is working kind of like a package manager for shell utilities / configuration, but it shouldn't really be a package manager. That's what you use `brew` or whatever it is you use for.

pkg/omf/cli/omf.info.fish
sheldon: good point, dropped it out

pkg/omf/omf.fish
derekstavis: We could use `d`, `de`, `describe`.

bucaran: Options

+ `d` for describe
+ `e` for explain
+ `a` for about

Or keep `info` and use `f`, and `inf` for the aliases.

bpinto:
I'll test it later but I like the idea!

You need to add this command to help.

pkg/omf/cli/omf.info.fish
sheldon: nice spot, the error handling was all wrong, I've updated it now to properly check for the package existing

pkg/omf/omf.fish
sheldon: sure, how about describe

sheldon:
@bucaran good point, I think the slow response isn't too bad anyway, and you can ask for 1 exact package directly for a quick output

@bpinto added in the help now

bucaran:
@sheldon And to completions! :smile: 

@bpinto The same goes for `cd`.

sheldon:
@bucaran @bpinto added in the completions, and cd too :smile:

bucaran:
@bpinto Maybe `d` is not a good idea? Because it can be interpreted as `delete`? 

@sheldon Sorry about the indecisiveness.

sheldon:
no problem, makes sense the clash

sheldon:
the other 2 suggestions of about or explain both work without the clash, any preferences?

bpinto:
Info or describe are the best. Regarding these short commands, I am still going evaluating it in my head. I honestly don't like and don't use any myself.

bpinto:
Thanks for the cd autocomplete but there is one more missing for both commands, the one that will complete the package/theme name.

sheldon:
added in showing the readme of the repo when describing just 1 package

sheldon:
@bpinto sure, I've added in the package name completions too

bpinto:
Full readme is unnecessary?
```
⋊> omf on sheldon-info-command ◦ omf describe ansible                                                                                                                     15:00:53
:black_circle: Ansible development package for Wahoo

[![](https://img.shields.io/badge/Wahoo-Package-00b0ff.svg?style=flat-square)][Wahoo]
![](https://img.shields.io/badge/License-MIT-707070.svg?style=flat-square)

# :black_circle: [_ansible_](https://github.com/ansible/ansible)
> Ansible is a powerful automation tool that you can learn quickly.

Ansible development for [Wahoo][Wahoo] / [fish](fishshell.com).

## Install
> _Note_: You need to have `ansible` source code handy and set `ANSIBLE_HOME` prior to loading Wahoo.
```

bpinto:
By the way, I like this very much! :D

sheldon:
sure, the full readme is a little overkill, I got carried away! :)

bpinto:
@sheldon 
Rebase so we can merge?
Squashing your commits would also be interesting.

sheldon:
sure, all in 1 commit now

derekstavis:
[LGTM](http://lgtm.in/i/owaBPNIcq)

bpinto:
Thanks @sheldon !

derekstavis:
Now you was faster, @bpinto

PR #551, opened 2015-08-29
registry: add foreign-env plugin

derekstavis:


bpinto:
You closed at the same time I hit merge... you won. :P

PR #555, opened 2015-08-29
Fix #545: Remove custom packages and themes

### LINKED ISSUE #545: `omf remove` is not deleting themes
Create a custom theme:
```
omf new theme custom-theme                                                                                                                         
```

Remove the theme
```
omf remove custom-theme
```

Expected: `~/.config/omf/themes/custom-theme` to be erased, but it's not.

bpinto:
Remove function was not deleting packages and themes from the OMF_CONFIG
folder.

### DIFF
--- a/pkg/omf/cli/omf.remove_package.fish
+++ b/pkg/omf/cli/omf.remove_package.fish
@@ -1,5 +1,7 @@
 function omf.remove_package
   for pkg in $argv
+    set -l remove_status 1
+
     if not omf.util_valid_package $pkg
       if test $pkg = "omf"
         echo (omf::err)"You can't remove `omf`"(omf::off) 1^&2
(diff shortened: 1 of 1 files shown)


derekstavis:
Tested here... And

[LGTM](http://lgtm.in/i/GTvEFPs6q)

bobthecow:
It probably should refuse to remove packages and themes inside `$OMF_CONFIG` because those are user data. We didn't put them there.

bpinto:
We did, when you do 'omf new theme abc' we add it there. That's why I missed this functionality so bad.

bobthecow:
I disagree. We made a skeleton for you. That's not the same. 

Removing something installed off of github is easily reversible. Just tell omf to install it again. There's no way at all to reverse removing something from inside $OMF_CONFIG.

bpinto:
What do you think about adding a confirmation if it's inside `$OFM_CONFIG`. My opinion is, if we were able to create inside $OFM_CONFIG we should also be able to delete. We can add a warning/confirmation though.

bucaran:
@bpinto I never implemented this feature because OMF_CONFIG and OMF_CUSTOM were separate entities before. Now that they were merged, I am not so sure this is a terrible idea, but this:

>  There's no way at all to reverse removing something from inside $OMF_CONFIG.

Adding a pretty Y/n confirmation box using `fish` will be a PITA, just try. If we allow ourselves to use bash for that bit then we could make it.

But instead of doing any of that we could just move the files to the trash. There are ways to make this work in both xdg and osx.

bpinto:
I like your trash idea! It could be enough, but as we need a confirmation for 'omf destroy' perhaps we should try implementing a confirmation with bash? Do you have any experience with it? I have never tried!

bucaran:
There are a few ways. It's a hideous hack. I'll look it up later then.

bpinto:
@bucaran thanks man!

PR #556, opened 2015-08-29
cli: add available command to list remote packages

derekstavis:
We need a way to display remote packages. `omf install` without parameters is counter-intuitive.
This pull request allows us to list remote packages using one of following commands:

    omf available
    omf av
    omf a

This is in preparation for using `omf install` as a dotfile resolver.


bpinto:
I wouldn't go this route... Can we add an option to -list instead?

List --remote
List --available
List --local (default)

bucaran:
See, this is not the original way I intended Wahoo to work. @derekstavis Why do you love so much about complexity?

One of the best things about the framework that I like is how themes work. Although the code is aware of which themes are remote and which are installed locally, that's all transparent to the user. You don't need to know which are here or which are there. You only need to `omf theme this` and `omf theme that` and OMF will do the correct thing.

With plugins I wanted to do the same, hence the lack of all these options. Originally there was not even a way to skip packages, it was simply based on what was inside the `pkg`, soon to be renamed, `plugin` directory.

**If you type `omf install`, OMF will show you what packages you can install, so that's effectively the same as list remote packages.**

bpinto:
I am not following your concerns. May you explain again what you don't like about listing remote packages?

bucaran:
This command calls `list_db_packages` which shows the full contents of the registry. But what you need to know is which packages are available skipping the ones you already have. This feature is OK, but it would fit better inside:

```
List --remote
List --available
List --local (default)
```

Now, if we do that, then the CLI will start getting really complex very soon, so we may need to work with `getopts` or maybe I can write a smaller 100 LOC `getopts` with the same UI, but a bit more limited.

bpinto:
I'd rather keep the CLI parser simple for now and continue using `switch` until we find a more complex command, which I agree with you, we should try to keep it simple (reason why I don't think `getopts` should be added).

bucaran:
:+1: Good

bpinto:
@derekstavis `list --remote` instead?

derekstavis:
Yes. I will give a look on this very soon.

PR #560, opened 2015-08-30
Do not convert `$HOME` when installing omf.

bpinto:
Keep `$HOME` variable inside `config.fish` when installing oh my fish
instead of converting it to its content.

### DIFF
--- a/bin/install
+++ b/bin/install
@@ -92,7 +92,8 @@ omf_install() {
   touch ${fish_config_file} >/dev/null 2>&1
   test ! -w ${fish_config_file} && die "Fish configuration file is not writable, aborting."
 
-  sed "s|{{OMF_PATH}}|$OMF_PATH|;s|{{OMF_CONFIG}}|$OMF_CONFIG|" \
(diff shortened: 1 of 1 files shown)


bpinto:
@bobthecow this should fix that problem you highlighted.

bucaran:
Which problem? I don't remember why anymore, but this was done in purpose, not just happened.

bpinto:
@bobthecow raise this.
> @bpinto the new (template based) install doesn’t use $HOME, which is far less multi-machine friendly.
bpinto 11:59

The existing code is new and it was created by my [here](https://github.com/fish-shell/oh-my-fish/commit/7c5e2b3964de3b6be18e37c264a16ee1efc50c4c), the proposed change is re-adding part of the code you did back to add `$HOME` as text to `config.fish` instead of evaluating the content of `$HOME` and adding it to `config.fish`.

bpinto:
I miss tests so much! :(

bucaran:
Exactly. Got it.

> I miss tests so much! :(

We'll get there.

bpinto:
@bucaran :+1: ?

bobthecow:
:+1:

PR #564, opened 2015-08-30
Gives back a way to persist and update installed packages (dotfiles)

derekstavis:
This PR makes omf cli install from and update a central file declaring installed plugins and themes.

The file is located at `$OMF_CONFIG/bundle`, and is a user-editable text file in the following format:

    package foreign-env
    theme agnoster copy

Blank lines are ignored as do lines that don't start with `package` or `theme`.

What it changes from current cli:

+ `omf install` (without parameters):
  - Looks for all packages declared in bundle file
  - Check what's missing on user installation
  - Install missing packages, satisfying the bundle
+ `omf install [package|theme ...]`
  - The install procedure is unchanged
  - If successful, add the installed package to the bundle (if not already there!)
+ `omf remove [package|theme ...]`
  - The remove procedure is unchanged
  - If successful, remove the installed package from the bundle (if already there!)

On clean installs where `$OMF_CONFIG` exists, the installer will satisfy the bundle before reloading back into fish.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,17 +81,16 @@ Oh My Fish includes a small utility `omf` to fetch and install new packages and 
 
 Update framework and installed packages.
 
-## `omf install` _`<package> ...`_
+## `omf install` _`[package|url ...]`_
 
-Install one _or more_ themes or packages. To list available packages type `omf theme`.
+Install one _or more_ packages.
 
-> You can fetch packages by URL as well via `omf install URL`
+- You can install packages directly by URL via `omf install URL`
+- When called without arguments, install missing packages from [bundles](#dotfiles).
 
 ## `omf list`
 
 List installed packages.
-
-> To list packages available for download use `omf install`.
 
 ## `omf theme` _`<theme>`_
 
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ Uninstall Oh My Fish. See [uninstall](#uninstall) for more information.
 
 # :triangular_flag_on_post: Advanced
 + [Startup](#startup)
++ [Dotfiles](#dotfiles)
 + [Core Library](#core-library)
 + [Packages](#packages)
   + [Creating](#creating)
@@ -137,6 +138,18 @@ Uninstall Oh My Fish. See [uninstall](#uninstall) for more information.
 This script runs each time a new session begins, autoloading packages, themes and your _config_ path in that order.
 
 The _config_ path (`~/.config/omf` by default) is defined by `$OMF_CONFIG` in `~/.config/fish/config.fish`. Modify this to load your own configuration, if you have any, as discussed in the [FAQ](FAQ.md#what-does-oh-my-fish-do-exactly).
+
+## Dotfiles
+
+The `$OMF_CONFIG` directory represents the user state of Oh My Fish, and is the perfect
+candidate for being added to your dotfiles and/or checked out to version control. There are two important files:
+
+- __`theme`__ - The current theme
+- __`bundle`__ - List of currently installed packages/themes
+
+### About the bundle
+
+Everytime a package/theme is installed or removed the `bundle` file is updated. You can also edit it manually and run `omf install` afterwards to satisfy the changes. Please note that while packages/themes added to the bundle gets automagically installed, a package/theme removed from bundle isn't removed from user installation.
 
 ## Core Library
 
--- a/bin/install
+++ b/bin/install
@@ -100,7 +100,10 @@ omf_install() {
     echo "Writing Oh My Fish configuration → ${OMF_CONFIG}"
     mkdir -p "${OMF_CONFIG}"
     touch "${OMF_CONFIG}/theme"
+    touch "${OMF_CONFIG}/bundle"
     test -f "${OMF_CONFIG}/revision"  || echo ${git_rev} > "${OMF_CONFIG}/revision"
+  else
+    fish -c "omf install"
   fi
 }
 
--- a/pkg/omf/cli/omf.help.fish
+++ b/pkg/omf/cli/omf.help.fish
@@ -2,10 +2,10 @@ function omf.help -a type_flag
   switch $type_flag
     case "i" "install" "get"
       echo \n"\
-      Install and list packages.
+      Install packages.
 
       "(omf::dim)"Usage:"(omf::off)"
-        omf install                 List available packages to install
+        omf install                 Install missing packages from bundle
         omf install "(omf::em)"<name>"(omf::off)" | "(omf::em)"<url>"(omf::off)"  Install package by name or URL
 
       "(omf::dim)"Examples:"(omf::off)"
--- a/pkg/omf/cli/omf.install.fish
+++ b/pkg/omf/cli/omf.install.fish
@@ -1,4 +1,26 @@
 function omf.install -a type_flag name_or_url
+  function _write_bundle
+    set bundle $OMF_CONFIG/bundle
+    set record $argv
+
+    if test -f $bundle
+      if not grep $record $bundle > /dev/null 2>&1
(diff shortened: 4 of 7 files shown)


bucaran:
You know, there just can't be races in the community. Don't get me wrong, if you _really_ want to do this feature, sure, go ahead and by all means, but I **assumed** I was going to take care of it. This is not a competition, but having believed the previous statement I went ahead and put a decent amount of work two days ago day despite being busy with real work, and unfortunately can't only really focus on this after a few days.

I am not even mad or disappointed, I'm just burned out of "fighting" over all this stuff /sigh. It's funny.

derekstavis:
@bucaran I've put a massive amount of work on this. It's working, it was exhaustively tested by me and it's commits are very logical and organized. I think we need to solve this problem ASAP, and implementing by myself something that __just works__ seemed the best thing to do.

bucaran:
@derekstavis There's no rush that can't wait a few days, not even a week and I'm on the same boat. 

Don't take it personally, I just thought it was funny how you went ahead and did what you wanted to do anyway.

sheldon:
Hey @derekstavis 

I kind of like that install without arguments offers what those arguments could be. Backwards compatibility goes a long way to keeping users happy. :smile: 

What about making update behave this way in addition to updating packages instead? It seems more natural to me that update would install missing packages from a registry.

As the output from update without arguments is unlikely to be parsed, whereas install without arguments is the primary way to list available packages currently, I think it's a safer change to previous behaviour too.

Another option: a sync command instead.

bpinto:
Until last week, omf install without arguments would install all packages on your config file, so I think this is actually solving backwards compatibility issues for 99% of the users.

My experience working with other CLI tools (e.g brew, bundler) also lead me to think that omf install would install many packages instead of listing the packages.

I don't know any tool where an update will actually call install, do you have an example?

sheldon:
Nothing comes to mind. Funnily enough, before last week I hadn't used this before :)

I'm all for sticking with the way things were before. Fixing regression ftw.

derekstavis:
`omf update --bundle` is what I was thinking for the future PR I'm planning on for version control.

bpinto:
LFTM

@bucaran @bobthecow opinions?

bucaran:
:+1:

scorphus:
I like this approach. And I'd add that I like the `--sync` option, but some may see it as superfluous, I guess. I also miss a way to list all the available packages.

derekstavis:
I've rebased it based on master @bpinto.
@scorphus I will follow with a PR for allowing a better `list` command.

scorphus:
:+1:

bobthecow:
If we're going to go with something simple, why not two files, `packages` and `themes` each of which have one entry per line with the name of the package or theme? Having something that we have to parse just seems like overkill for this. Having a third concept (themes, packages and now bundles) seems like overkill as well.

derekstavis:
Great point @bobthecow. The problem I had with `packages` and `themes` approach is that currently we can have directories with this names inside `$OMF_CONFIG`. As we are going to rename `package` to `plugin` the same could happens. That's why I preferred a single file rather than different files, which could make more sense, and avoid the need to parse it (although the parser is very simple).

bpinto:
I like the current concept of `bundle` and `theme`. They have different purposes.

`bundle`: Themes and packages that you have installed and loaded.
`theme`: Current loaded theme.

We might want to remove the `theme` file in the future, but it might be a bigger discussion.

derekstavis:
I like the fact that the bundle can contain more than one theme. I often swap themes on my different machines during the day. Sometimes the theme is too heavy for my little embedded machines, and while on heavy terminal usage I swap to simple theme (recalculating git every enter isn't great on them) and when I need git stuff I change to `agnoster` of `bobthefish`.

bobthecow:
Wasn't there one of the pull requests floating around that moved the theme and revision files into a subdirectory to make it more obvious that it was a programmatically generated one rather than a user file? Seems that would solve this issue.

bpinto:
Hum... let me think:

We have three files:

- `revision` - not used, can be removed.
- `theme` - saves current theme, might also be removed in the future.
- `bundle` - saves themes and packages

I don't know, but I'd rather try to remove both `revision` and `theme` file instead of moving all three to a new directory.

derekstavis:
`theme` file is OK for me for dotfiles usage. It's great to clean install `omf` keeping `$OMF_CONFIG` and see how everything gets back as it was.

derekstavis:
I just rebased this PR to master for anyone who wants to test it.

scorphus:
@bucaran Did you have time to test it?

@all What are the major issues you see with this approach?

scorphus:
@derekstavis What if I already have a bunch of packages installed, like is the case for most of us?

This is what I did:

```fish
omf.list_installed_packages | awk '{print "package " $1}' > $OMF_CONFIG/bundle
omf.list_installed_themes | awk '{print "theme " $1}' >> $OMF_CONFIG/bundle
```

derekstavis:
Yes, this is a easy way to sync current state! :+1:

scorphus:
But that's just a *makeshift-that-solved-my-problem* thing.

scorphus:
I actually expected either `omf install` or `omf update` would do this favor to me.

bucaran:
Not yet. I will later today. 

Another suggestion, what about just keeping a file with the plugin names and be done with this?

Users can then omf install (cat $OMF_CONFIG/plugins)

derekstavis:
I think we don't want to go `omf install (cat $OMF_CONFIG/plugins)` way. This is such a bad cli. About keeping just a file with plugins forgets about theme support.

bucaran:
There is no multiple theme support in the traditional sense. OMF installs your theme with omf theme name whenever you like. There is no need to keep a list of local and remote themes.

bucaran:
TBH I don't think the CLI is bad. This is shell business. If we wanted a more sophisticated CLI then we would use a higher level language.

bucaran:
I love the idea of simply omf theme name and use the theme I want without knowing whether the theme is already installed or not. It just works. We want a fast framework that gets out of the way right?

derekstavis:
>I love the idea of simply `omf theme name`

So simple that it tries to upgrade the theme everytime you change it.

We want a fast framework that gets out of the way right?

>I think everyone wants a flexible and extensive framework done by everyone not just by one.

bucaran:
We could remove the autoupdate behavior.

Sure. IMO this PR adds too much complexity. I want to simplify, remove files, merge commands, reduce the cognitive load imposed in the user so that anyone can use OMF without having to read any docs if they feel lazy.

derekstavis:
Sorry. Your argument is really invalid. Look back to your PR again and then come and say this one adds complexity. The problem isn't with the PR, it's with your ego. You need to accept people ideas, not just want to impose yours and assume the project is yours (even it being now). This kind of behaviour is toxic, and that's why now omf is broken. And don't come talk to me about user experience when you were the first that used your persuasion and done a big mistake by letting all omf users fuck themselves.

bucaran:
Why? The only main feature missing in Wahoo was what you are pushing for here and which I find mostly (not completely) useless.

omf install (cat $OMF_CONFIG/plugins)
 ...or we could add a command that does that.

In fact, we could easily allow users to add their own commands to omf (or just write a plugin for that).

Even better, let's break omf commands into subplugins and make core even simpler.

bpinto:
@derekstavis
Take it easy... We are again back to the same unrelated comments...

@all
Things are not moving forward, most agreed this was good. So I'm using that to merge it. End of story.

bucaran:
@bpinto This move was too fast.

bpinto:
Topic is closed, please let's move on.

bucaran:
I will give it a shot.

PR #567, opened 2015-08-30
describe works on installed packages

### LINKED ISSUE #563: `omf describe` only describes uninstalled packages
`omf describe` uses `omf.list_db_packages` to get the list of packages, but this removes installed packages from the list of packages.

IMO, this function by default should not skip installed packages and receive an argument that will do this instead.

sheldon:
also modified list_db_packages to return all packages by default, but with a
skip_installed to skip listing installed packages

fixes #563

### DIFF
--- a/pkg/omf/cli/omf.list_db_packages.fish
+++ b/pkg/omf/cli/omf.list_db_packages.fish
@@ -1,6 +1,11 @@
 # List all packages available to install from the registry.
-function omf.list_db_packages
+function omf.list_db_packages -a skip_installed
   for item in (basename $OMF_PATH/db/pkg/*)
-    contains $item (basename {$OMF_PATH,$OMF_CONFIG}/pkg/*); or echo $item
+    if  begin
+          test -z $skip_installed
+          or not contains $item (basename {$OMF_PATH,$OMF_CONFIG}/pkg/*)
+        end
+      echo $item
+    end
   end
 end
--- a/pkg/omf/completions/omf.fish
+++ b/pkg/omf/completions/omf.fish
@@ -16,7 +16,7 @@ complete -c omf -n "__omf.opt_is q query"         -a (printf "%s " (set | awk '{
 complete -c omf -n "__omf.opt_is r rm remove"     -a (printf "%s " (omf.list_local_packages) (omf.list_installed_themes))
 complete -c omf -n "__omf.opt_is d desc describe" -a (printf "%s " (omf.list_db_packages))
 complete -c omf -n "__omf.opt_is c cd"            -a (printf "%s " (omf.list_db_packages))
-complete -c omf -n "__omf.opt_is i install"       -a (printf "%s " (omf.list_db_packages))
+complete -c omf -n "__omf.opt_is i install"       -a (printf "%s " (omf.list_db_packages "skip installed packages"))
 complete -c omf -n "__omf.opt_is t theme"         -a (printf "%s " (omf.list_themes))
 complete -c omf -n "__omf.opt_is help"            -a "install theme remove update list describe cd new submit query destroy"
 
--- a/pkg/omf/omf.fish
+++ b/pkg/omf/omf.fish
@@ -89,7 +89,7 @@ function omf -d "Oh My Fish"
 
     case "i" "install" "get"
       if test (count $argv) -eq 1
-        omf.list_db_packages | column
+        omf.list_db_packages "skip installed packages" | column
       else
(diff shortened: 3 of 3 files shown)


pkg/omf/cli/omf.list_db_packages.fish
```diff
@@ -1,6 +1,11 @@
 # List all packages available to install from the registry.
-function omf.list_db_packages
+function omf.list_db_packages -a skip_installed
   for item in (basename $OMF_PATH/db/pkg/*)
-    contains $item (basename {$OMF_PATH,$OMF_CONFIG}/pkg/*); or echo $item
+    if  begin
```
bpinto:
```
if contains $item (basename {$OMF_PATH,$OMF_CONFIG}/pkg/*); or test -z $skip_installed
  echo $item
end
```
?

pkg/omf/completions/omf.fish
```diff
@@ -16,7 +16,7 @@ complete -c omf -n "__omf.opt_is q query"         -a (printf "%s " (set | awk '{
 complete -c omf -n "__omf.opt_is r rm remove"     -a (printf "%s " (omf.list_local_packages) (omf.list_installed_themes))
 complete -c omf -n "__omf.opt_is d desc describe" -a (printf "%s " (omf.list_db_packages))
 complete -c omf -n "__omf.opt_is c cd"            -a (printf "%s " (omf.list_db_packages))
-complete -c omf -n "__omf.opt_is i install"       -a (printf "%s " (omf.list_db_packages))
+complete -c omf -n "__omf.opt_is i install"       -a (printf "%s " (omf.list_db_packages "skip installed packages"))
```
bpinto: `omf.list_db_packages --skip-installed` ?

pkg/omf/omf.fish
```diff
@@ -89,7 +89,7 @@ function omf -d "Oh My Fish"
 
     case "i" "install" "get"
       if test (count $argv) -eq 1
-        omf.list_db_packages | column
+        omf.list_db_packages "skip installed packages" | column
```
bpinto: `omf.list_db_packages --skip-installed` ?

pkg/omf/cli/omf.list_db_packages.fish
sheldon: I thought so too, but it turns out ```or; ``` doesn't work in conditionals without a block. :(

bucaran: @sheldon Maybe I can help. 

Can you write in plain old C or Ruby the expression you are trying to achieve? I'll rewrite it in Fish.

sheldon: Sure, pretty much exactly what @bpinto wrote. The crux of the problem is fish's lack of && or || and ```; or``` and ```; and``` not working in if conditionals.

Here's the block that doesn't work in fish:
```fish
if test -z $skip_installed; or not contains $item (basename {$OMF_PATH,$OMF_CONFIG}/pkg/*)
  echo $item
end
```

my equivalent C-like psuedocode would be
```
if( ! skip_installed || item not installed){
  print item
}
```

sheldon: forgot to include the link with the syntax help I found: http://sourceforge.net/p/fish/mailman/message/31781593/

bucaran: There are workarounds, like an actual `&&` implementation written by one of fish core contributors, can't find the link now, but if you look you'll find it. Now, since this is fish anyway, what I prefer to do is:

```fish
if test -z "$skip_installed"
  if not contains $item (basename {$OMF_PATH,$OMF_CONFIG}/pkg/*)
    echo $item
  end
end
```

But I think you can also do this:

```fish
test -z "$skip_installed"; and contains $item (basename {$OMF_PATH,$OMF_CONFIG}/pkg/*); or begin
    echo $item
end
```

sheldon: I was struggling with trying to find a way to do this with nested ifs without repeating the echo. With a nested if we get the same as an && which isn't the right logic here. I think this is the best I could think ends up repeating the echo part:

```
if test -z "$skip_installed"
  echo $item
else
  if not contains $item (basename {$OMF_PATH,$OMF_CONFIG}/pkg/*)
    echo $item
  end
end
```

bpinto:
Thanks @sheldon

PR #579, opened 2015-09-01
Make `default` theme the default theme for omf.

bpinto:
* On clean install, configure and download `default` theme.
* Do not allow `default` theme to be removed.
* When the current theme is removed, fallback to `default`.


bucaran:
:+1:

bobthecow:
:+1:

PR #587, opened 2015-09-01
iterm2 shell integration

captdan:


bpinto:
I use both intern and fish but I don't know what this plugin does, could you add more information? I'd like to know what I am missing! :)

bobthecow:
@bpinto you use intern?

you should upgrade to intern2.

captdan:
I just realized there's a bit of a bug in its current form. I'll fix that up and add a description and resubmit 



———
Dan

bpinto:
@bobthecow LOL thanks for autocorrection! :D 

@bobthecow @captdan I thought iterm2 had fish support builtin.

bucaran:
Hi @captdan What's the status on this?

faho:
I don't use iTerm2 but this seems rather hacky - moving fish_prompt means you can't change it on-the-fly anymore. The code is also rather unfishy, with all the `x"$TERM" != "xscreen"` and triggers even if the term isn't iTerm2, so it's up to the user to enable it only then. It also only seems to trigger with an OMF theme.

I'd rather see what needs to be done in fish-shell to properly support it - do we need additional events to bind iterm2_pre{cmd,exec} and iterm2_prompt_{start,end} to?

bucaran:
`x"$TERM" != "xscreen"` looks like a bash hack, but what bothers me is the use of `[ ]` instead of `test`, but those are just pet peeves that can be worked on. Other than that I can't comment.

> I'd rather see what needs to be done in fish-shell to properly support it - 

This.

captdan:
https://iterm2.com/shell_integration.html

I'm also curious about how this could be supported by fish shell. the iTerm2 solution seems like a bit of a hack

bucaran:
@captdan The Fin repo is still not ready, so this PR belongs to https://github.com/wa/wahoo or https://github.com/oh-my-fish/oh-my-fish.

PR #592, opened 2015-09-03
Fix shell: sh -> bash

chessman:
Install script doesn't work in dash which doesn't have pushd.


bucaran:
Thanks! The script will be converted to fish shortly, so we will be able to avoid all these kind of issues in the future.

